PR #7, opened 2015-07-16
Remove VC6 hacks in AssignFromHelperClass.

### LINKED ISSUE #5: AssignFromHelperClass is horrible
The uses of Hack_DefaultValueFromConstReferenceType and Hack_GetValueIntoConstReference are undefined behaviour, and quite foul.

If the VC60 workaround is still needed (VC6 in 2015, really?) the sane way to do it would be:

	typename decay<R>::type value;
	if (!m_source.GetValue(name, value))
		throw InvalidArgument(std::string(typeid(T).name()) + ": Missing required parameter '" + name + "'");
	(m_pObject->*pm)(value);

with the following added as private members of `AssignFromHelperClass`:

	template <class R> struct decay { typedef R type; };
	template <class R> struct decay<const R&> { typedef R type; };

If you really must support VC6 then do it with `#if` not by using the same horrible code for compilers that aren't totally broken.

jwakely:
Fixes #5.

### DIFF
--- a/algparam.h
+++ b/algparam.h
@@ -171,18 +171,6 @@ GetValueHelperClass<T, T> GetValueHelper(const T *pObject, const char *name, con
 
 // ********************************************************
 
-template <class R>
-R Hack_DefaultValueFromConstReferenceType(const R &)
-{
-	return R();
-}
-
-template <class R>
-bool Hack_GetValueIntoConstReference(const NameValuePairs &source, const char *name, const R &value)
-{
-	return source.GetValue(name, const_cast<R &>(value));
-}
-
 template <class T, class BASE>
 class AssignFromHelperClass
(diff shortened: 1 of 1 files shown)


dirvine:
:+1:  This should be tested on the minimum required platform now and those platforms should be documented. I would suggest a commit, prior to merging this, in the Makefile/vcproj files to check compiler versions and error on unsupported platforms with a clear message at compile time.

DevJPM:
Sounds reasonable.

But before we can do this (either in Makefile/vcproj/config.h) we need
to definitely agree on a minimal platform, which we haven't done yet (if
I'm following things right).

BR

JPM

Am 16.07.2015 um 13:34 schrieb David Irvine:

PR #35, opened 2015-10-01
Fixes the warning on IntToString about comparison of unsigned and sig…

bertothunder:
This issue is mentioned somewhere else, but I found it while compiling a small program with crypto++; I'm using boost::uint32_t and boost::uint64_t with IntToString() method, in which case the internal comparison (a < 0) makes no sense.

The patch I'm pulling fixes this warning as the template checks the signedness of the input value; if the value is signed, the right comparison function is used, and when the input is unsigned, then returns false.

Haven't tried on Windows compilers, it works fine with gcc (since 4.4) and MacOSX.

### DIFF
--- a/misc.h
+++ b/misc.h
@@ -6,6 +6,8 @@
 #include "stdcpp.h"
 #include "trap.h"
 
+#include <limits>
+
 #ifdef _MSC_VER
 	#if _MSC_VER >= 1400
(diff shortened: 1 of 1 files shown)


mouse07410:
Haven't tested this PR yet, but it makes sense.

Issue #45, opened 2015-10-25
Attempt to support GCC's -fdevirtualize


noloader:
GCC offers `-fdevirtualize` option, which appears to be similar to Microsoft's `__declspec(novtable)`. See [Options That Control Optimization](http://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html)

Zireael-N:
I typed a long post about my observations, but then I had an idea, that basically was right. Made a pull request

---
> Third, I've seen problems with inlining functions. I had to move a few functions out-of-line because of crashes. I can't explain why it avoids the crash.

That was the culprit.

---
> Fourth, I've never tried `-fdevirtualize`

It's [a part of -O2](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html). :)

noloader:
>> Fourth, I've never tried -fdevirtualize
> It's a part of -O2. :)

OK< thanks. That's good to know.

I'm trying to track down a rare crash in SHA. The only person experiencing it is a Debian maintainer who regularly uses Sid. I've tried to duplicate it but I cannot. I think I might ask him to try `-fno-devirtualize`.

noloader:
Zireael - how do you want us to give you credit in the Readme and source files? Is "Zireael-N N" OK, or do you want something else?

By the way, your fix cleared that obscure hang on Debian that I was chasing for weeks.

Zireael-N:
> By the way, your fix cleared that obscure hang on Debian that I was chasing for weeks.

Glad to hear that!

> Is "Zireael N" OK, or do you want something else?

Just "Zireael" is enough.

I've also updated the bug report on GCC's bugzilla. Hopefully, they'll look into that.

noloader:
>> By the way, your fix cleared that obscure hang on Debian that I was chasing for weeks.
> Glad to hear that!

I got confirmation from the Debian folks today. It cleared the issue.

Thanks again. I owe you a cold beer if you make it to my area. Baltimore MD/Washington DC or New York.

Zireael-N:
Haha, I wish I could! I'm not from the US.

PR #46, opened 2015-10-26
Remove implementation of ResizeBuffer() from headers

Zireael-N:
Fixes issue #45

### DIFF
--- a/modes.cpp
+++ b/modes.cpp
@@ -149,6 +149,12 @@ void BlockOrientedCipherModeBase::UncheckedSetKey(const byte *key, unsigned int 
 	}
 }
 
+void BlockOrientedCipherModeBase::ResizeBuffers()
+{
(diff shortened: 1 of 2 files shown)


noloader:
Closing...

noloader:
Thanks. Tracking these down is a formidable task at times...

noloader:
For completeness, we reworked all three calls to `ResizeBuffer` because of our knowledge of what the optimizing the call to `New`.

Here's what the code looks like for 5.6.3:

    // Thanks to Zireael N, http://github.com/weidai11/cryptopp/pull/46
    #ifndef CRYPTOPP_MAINTAIN_BACKWARDS_COMPATIBILITY_562
	virtual void ResizeBuffers();
    #else
	virtual void ResizeBuffers()
	{
		m_register.New(m_cipher->BlockSize());
	}
    #endif

PR #65, opened 2015-11-08
Linux shared-object library versioning

tknarr:
Changes to do shared-object library versioning on Linux. The soname is based on the major and minor version numbers, and presumes patchlevel updates are safe to use with binaries compiled and linked against older patchlevels. Mainly that means header files don't change significantly across patchlevels. If that's not the case, the soname would have to have the patchlevel added to it and the link creation/removal code involving SOLIB_COMPAT_SUFFIX has to change.


noloader:
Thanks @tknarr . We need this. I'll move the discussion to the list to ensure no one objects.

A quick question (if you don't mind)... Line 367:

    +      ldconfig

Is it OK for us to call that on install? I recall trying to find provenance for it in Stallman's GNU Make book, but there was no treatment.

tknarr:
ldconfig isn't in GNU Make, it's part of the Linux linker/loader package. When you install or uninstall a shared library, ldconfig updates the symlinks and /etc/ld.so.cache so the loader can find the library at runtime. It may be different on other OSes, that's why I only did versioning for Linux. An alternative is 'ldconfig -n $(PREFIX)/lib' which would just update the symlinks in the lib install directory, but without updating the cache the loader may not be able to find the library.

The bit about it I don't like is that it fails when you run ldconfig as a normal user. I'm not sure of a way around that.

noloader:
Todd - we are deferring this until 5.7. We should be able to merge it in a couple of weeks.

tknarr:
OK. And if anyone's twitchy about ldconfig I can change that to just do the symlinks and leave rebuilding the cache for the user to do afterwards.

noloader:
@Todd,

> And if anyone's twitchy about ldconfig I can change that to just do the symlinks and leave rebuilding the cache for the user to do afterwards

Yes, let's do that. That is, use SO name, use symlink but leave ot to the user to rebuild the cache.

I think its OK to INFO the user, similar to what the makefile does for its public service announcement:

    $ make
    WARNING: CRYPTOPP_NO_UNALIGNED_DATA_ACCESS is not defined in config.h.
    WARNING: You should make these changes in config.h, and not CXXFLAGS.
    WARNING: You can 'mv config.recommend config.h', but it breaks versioning.
    WARNING: See http://cryptopp.com/wiki/config.h for more details.

Instead of a WARN, let the user know they may need to do it at an INFO level. I can only say *MAY* because I know its not needed on Fedora and Ubuntu when installing into `/usr/local`. But I also know its needed on OpenSUSE and possibly Gentoo.

tknarr:
The info message sounds good. I'll verify that I can use the links-only command as a non-root user.

The code already does the 3-tuple, if you were looking at SOLIB_COMPAT_SUFFIX that's supposed to only be 2 elements so you can fix bugs and automatically have programs linked against that minor version pick up the fix without needing to be relinked.

noloader:
> I'll verify that I can use the links-only command as a non-root user.

OK, sounds good. Thanks.

I think the two use cases are:

 1. Build in PWD, don't install, test in PWD
 2. Build in PWD, install into `/usr` or `/usr/local`, test from anywhere.

I use (1) frequently. When I link, I link to the static archive with `./libcryptopp.a`.

I use (2) on occasion. I perform a `sudo make install PREFIX=/usr/local`. When I use (2), I compile/link test programs with `-L/usr/local/lib` and `-lcryptopp`. I perform a `LD_LIBRARY_PATH=/usr/local/lib ./myprog` if needed. But its usually not needed on platforms like Fedora and Ubuntu.

I've never heard of anyone performing a `make install PREFIX=/"$HOME"`, but it could be a valid use case.

I'm not sure about what others might do. But I can speak for the two use cases above.

cawka:
I would like to propose a few changes here.  For C++, the concept of compatibility between major releases is not that straightforward.  Any minor change in class interface can cause library to be incompatible with previous version.

So.  I'm supporting versioning, but I would like to use full version as a soname:  libcryptopp.so.5.6.2  (and libcryptopp.5.6.2.dylib for os x).

GamePad64:
> Any minor change in class interface can cause library to be incompatible with previous version

In fact, that is not really true. I know about [Fragile ABI](http://c2.com/cgi/wiki?FragileBinaryInterfaceProblem), but there are ways, to make ABI compatible between versions. I strongly oppose this change, because every minor library update will result in rebuilding of all binaries, dependent on this library.

noloader:
Hi Guys. So everyone is on the same page, we have this wiki page setup: [Release Versioning](https://cryptopp.com/wiki/Release_Versioning).

Here's the three second tour:

* View from a `dlopen` and `dladdr` perspective
** attempts to consider/convey if/when a client must recompile
** i.e., level of "breaking" with respect to breaking changes

* Revision version bump (x.y.Z)
** maintain the library
** ideally, nothing is disturbed
** new classes on occasion
** no exiting API changes
** no ABI changes

* Minor version bump (x.Y)
** advance the library
** new classes added
** preserve exiting order of vtables entries
** may add vtable entries

* Major version bump (X)
** update the library
** breaking changes
** clients probably recompile
** classes added, removed, modified
** API and Interfaces added, removed, modified
** ABI and vtables added, removed, modified

tknarr:
Compatibility's why I used SOLIB_COMPAT_SUFFIX in GNUmakefile. To have only the exact version be considered compatible, just change SOLIB_FLAGS to use SOLIB_VERSION_SUFFIX and then comment out or delete the recipe lines that use SOLIB_COMPAT_SUFFIX. It only matters when a revision number bump would change existing code/data in header files to fix a bug, and when that happens you have to recompile applications anyway to get the fix.

cawka:
I agree and withdraw my suggestion.  Hopefully, the ABI compatibility can be achieved within minor/patch releases, though it is really hard to do with C++.

Ideally, this also requires additional testing, e.g., running the tester compiled with the old version against the new version of the shared library.

GamePad64:
You can try this tool for checking ABI compatibility:
http://ispras.linuxbase.org/index.php/ABI_compliance_checker
I tried it on other project, but didn't succeed very much. Maybe, you will
be more lucky with it.

2015-12-15 23:21 GMT+03:00 Alex Afanasyev <notifications@github.com>:

noloader:
@tknarr - I managed to break your patch, and I cannot merge it. Would you be able to fix it one last time?

tknarr:
On 12/25/2015 09:14 AM, Jeffrey Walton wrote:
>
> @tknarr <https://github.com/tknarr> - I managed to break your patch, 
> and I cannot merge it. Would you be able to fix it one last time?
>
Sure thing.

tknarr:
That should do it.

noloader:
Thanks @tTodd, got it.

I'm going to start testing with it, so I may be pining'ing you for some refinements.

noloader:
Hey Todd. Sorry to ping you again. Under Cygwin we are catching:

```
make: No rule to make target 'libcryptopp', needed by 'libcryptopp.so'. Stop
```

I can't test MinGW, but there's a couple of folks we can ping and ask them to test. The first is @Zireael-N and the second is @IlyaBizyaev.

tknarr:
Found it. It's in the line to allow libcryptopp.so as a target when versioned. I forgot the .so on the right-hand side, and I forgot I need to make that target conditional so it's not present when we aren't versioned. Fixing that now, pull request incoming shortly.

Issue #73, opened 2015-12-07
Provide CMake build system


GamePad64:
Well, the idea is to move away from traditional Makefile to a Makefile generator. CMake is the most robust, stable and reliable cross-platform build system.

Advantages:
- Much, MUCH easier to maintain:
Now we have to mantain Visual Studio project, Borlang C++ Builder project, GNUmakefile, GNUmakefile-cross, and (bonus!) Debian maintainers also have to maintain autoconf/automake builds. Using CMake you just have to maintain a single file: CMakeLists.txt
- Uniform across all platforms
We don't need to remember, how to build Crypto++ on Windows or Android (and it is really hard without something like msys2, if using MinGW).  
Building process is the same on every single platform:
```shell
mkdir build
cd build
cmake ../build
cmake --build .
```
- Cross-platform out of the box:
CMake uses [toolchain files](https://cmake.org/cmake/help/v3.0/manual/cmake-toolchains.7.html) to be really crossplatform. If you want to cross-compile Crypto++ for Android or Raspberry PI, for example, you just create CMake toolchain file for each platform, and it works like magic.
- Great documentation and community
CMake is becoming *de facto* standard in building C and C++ applications, gaining popularity every day.
- Many IDE's start to support it out of the box:  
Visual Studio (using generator)  
Code::Blocks (using generator)  
Eclipse CDT (using generator)  
CLion (natively!)

Disadvantages:
- It would be necessary to rewrite building process in various Linux distros.

If you are agree with this idea, I can start writing the CMakeLists.txt file right now, so we could complete it until version 5.7. Also, we can roll out both Makefile+CMake builds simultaneously in 5.7, and remove Makefile in 6.0 (as it is a major release).

noloader:
> If you are agree with this idea, I can start writing the CMakeLists.txt 

Yes, I've been looking for a `CmakeList.txt` for some time. See, for example, [Add Android.mk to build using android NDK](https://github.com/weidai11/cryptopp/pull/3).

When you build your file list, you can perform the following to ensure you get the list the makefile uses:

    cd cryptopp-git
    git pull # recent check-in
    make sources
    Library sources: cryptlib.cpp cpu.cpp 3way.cpp adler32.cpp algebra.cpp...
    Test sources: bench.cpp bench2.cpp test.cpp validat1.cpp validat2.cpp...

[...]

PR #77, opened 2015-12-08
CMake build

GamePad64:
Hello,
I have added CMake build, as mentioned in #73.

Building static+shared library.
---------------------------------------
- Honors object linking order, to contend [Static Initialization Order Fiasco](https://cryptopp.com/wiki/Static_Initialization_Order_Fiasco).  
- Forces adding -fPIC to all 64-bit architectures, such as amd64 and aarch64. For 32-bit architectures behaviour is controlled using `-DCMAKE_POSITION_INDEPENDENT_CODE=ON/OFF`
- Warns if `CRYPTOPP_NO_UNALIGNED_DATA_ACCESS` or `CRYPTOPP_INIT_PRIORITY` not defined in config.h
- Detects and sets endianess.
- `DISABLE_ASM`, `DISABLE_SSSE3` and `DISABLE_AESNI` are **not** set automatically right now (using compiler detection), so they are made as CMake options.
- Links to ws2_32 on Windows and to pthread on Linux.
- SONAME support for Linux (version_major.version_minor)

Building tests (cryptest).
--------------------------------
The test can be built and performes, if `BUILD_TESTING` is set to `ON` (default).
Command: `make test`

Building documentation using Doxygen.
-----------------------------------------------------
I had to create a new file, `Doxyfile.in`, as Doxygen can't make out-of-source builds. This is *the right way* to do this in CMake, but we may have to maintain it :confused:
The build yields CryptoPPRef.zip, containing all documentation, generated by Doxygen (just like GNUmakefile build).

Install.
--------
On `make install` CMake installs a package, that could be used by other projects, using `FindPackage(cryptopp)`, eliminating the need of `FindCryptoPP.cmake` module.
Also, `make install` just installs the libraries in `lib`, includes in `include`, and documentation in `share/doc`, honoring FHS on Linux.

Also, I've added a working `.travis.yml` file for [Travis-CI](https://travis-ci.org), if you ever want to use it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,185 @@
+cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+project(cryptopp VERSION 5.6.3)
+
+include(GNUInstallDirs)
+include(TestBigEndian)
+include(CheckCXXSymbolExists)
+
+#============================================================================
+# Settable options
+#============================================================================
+
+option(BUILD_TESTING "Build library tests" ON)
+option(BUILD_DOCUMENTATION "Use Doxygen to create the HTML based API documentation" OFF)
+
+option(DISABLE_ASM "Disable ASM" OFF)
+option(DISABLE_SSSE3 "Disable SSSE3" OFF)
+option(DISABLE_AESNI "Disable AES-NI" OFF)
+
+#============================================================================
+# Internal compiler options
+#============================================================================
+
+set(LIB_VER ${cryptopp_VERSION_MAJOR}${cryptopp_VERSION_MINOR}${cryptopp_VERSION_PATCH})
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "Intel")
+	add_definitions(-wd68 -wd186 -wd279 -wd327 -wd161 -wd3180)
+endif()
+
+# Endianess
+TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
+if(IS_BIG_ENDIAN)
+	add_definitions(-DIS_BIG_ENDIAN)
+endif()
+
+if(DISABLE_ASM)
+	add_definitions(-DCRYPTOPP_DISABLE_ASM)
+endif()
+if(DISABLE_SSSE3)
+	add_definitions(-DCRYPTOPP_DISABLE_SSSE3)
+endif()
+if(DISABLE_AESNI)
+	add_definitions(-DCRYPTOPP_DISABLE_AESNI)
+endif()
+
+#============================================================================
+# Sources & headers
+#============================================================================
+
+# Library headers
+file(GLOB cryptopp_HEADERS *.h)
+
+# Test sources
+file(GLOB cryptopp_SOURCES_TEST bench.cpp bench2.cpp test.cpp validat1.cpp validat2.cpp validat3.cpp adhoc.cpp datatest.cpp regtest.cpp fipsalgt.cpp dlltest.cpp fipstest.cpp)
+
+# Library sources
+file(GLOB cryptopp_SOURCES *.cpp)
+list(REMOVE_ITEM cryptopp_SOURCES
+		${CMAKE_CURRENT_SOURCE_DIR}/cryptlib.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/cpu.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/pch.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/simple.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/winpipes.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/cryptlib_bds.cpp
+		${cryptopp_SOURCES_TEST}
+		)
+set(cryptopp_SOURCES
+		${CMAKE_CURRENT_SOURCE_DIR}/cryptlib.cpp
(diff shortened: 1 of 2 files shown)


noloader:
This is a very good start. Thank you very much.

******

CMakeLists.txt, Lines 19-31, can go away. Future versions of the library will apply `CRYPTOPP_NO_UNALIGNED_DATA_ACCESS` and `CRYPTOPP_INIT_PRIORITY`. "Future versions" mean the next release (which may be 5.7 or 6.0, depending on what changes).

Folks who *want* unaligned data access will have to specifically set `CRYPTOPP_ALLOW_UNALIGNED_DATA_ACCESS`. That means [confg.h block around line 500](https://github.com/weidai11/cryptopp/blob/master/config.h#L497) will be deleted. We will no longer set it automatically.

******

I'm not sure about CMakeLists.txt, Lines 120-125. This is due to my ignorance.

    +if(WIN32)
    +	target_link_libraries(cryptopp-static ws2_32)
    +	target_link_libraries(cryptopp-shared ws2_32)
    +endif()

So we are clear... I believe the intention in the GNUmakefile is Unix-on-Windows tools like Cygwin and MinGW. MinGW requires the Winsock library `ws2_32`, while Cygwin does not. I don't believe its a cross-compile case, like using GCC to produce PE executables from Ubuntu under WINE (or something else exotic).

This is something you may want to test in a VM to be sure.

******

I'm not sure about CMakeLists.txt, Lines 125-130. 

    +find_package(Threads)
    +target_link_libraries(cryptopp-static ${CMAKE_THREAD_LIBS_INIT})
    +target_link_libraries(cryptopp-shared ${CMAKE_THREAD_LIBS_INIT})

On OS X, there are no thread libraries used. See [Crypto++ build on OS X 10.8](http://pastebin.com/uJS8xWz8) (it expires in 30 days).

*****

You might consider dropping Doxygen support. 5.6.2 and earlier did not have the make-based support. I believe Wei was generating it by hand in a one-off fashion.

We added the make-based stuff at 5.6.3 to ensure consistency are reproducible results. It was a QA thing.

If I did not need to generate the docs, then I'd avoid it altogether. (I think you have this choice).

******

travis.yaml concerns me.

    + sources: ['ubuntu-toolchain-r-test', 'george-edison55-precise-backports']

Effectively, we are introducing external dependencies. In addition, we are asking folks to trust someone named George Edison 55.

Can they be removed?

******

I'm not sure about this in travis.yaml:

    + - CXX_COMPILER=g++-5 STDLIB=libc++

The library supports a wide array or compilers, including MSVC, GCC, Intel ICC, Sun Studio, etc. I'm not sure limiting to GCC 5+ is a good idea.

In fact, I've seen *a lot* of problems with the experimental compilers like 5.2 and 5.3. I know they have inlining bugs, and I have spent a couple hundred hours trying to track them down on platforms like Debian/Aarch64. `libc++` is LLVM's C++ runtime. Is there a reason it is specified, and not, say, GNU's `libstdc++` or STLport or Dinkumware?

My apologies if I am parsing things incorrectly.

******

`Doxygen.in` looks like its a copy of Doxygen. Does a pre-build step of `cp Doxygen Doxygen.in` work to meet the goal?

Copying `Doxygen` has the benefit of no `Doxygen.in` maintenance because we will tend to it. Doxygen's configuration file will always be maintained because we use it to build the docs.

******

I think that's enough for round one.

I'll get it into the testing loop, and then the real fun will begin. (Testing sucks at times. For every half hour I spend producing something, I spend hours testing it).

noloader:
This is now up for debate at [CmakeList.txt Pull Request](http://groups.google.com/d/msg/cryptopp-users/_CRpt7-E20k/2urMb6A7CQAJ).

GamePad64:
> CMakeLists.txt, Lines 19-31, can go away.
Fair point, done
> I'm not sure about CMakeLists.txt, Lines 120-125.
WIN32 is not defined in Cygwin using CMake version higher then (2.8.4). CMakeLists.txt requires 3.2 explicitly.
> I'm not sure about CMakeLists.txt, Lines 125-130.
`CMAKE_THREAD_LIBS_INIT` is defined only when needed (on Linux, specifically). On OS X it is no-op.
> travis.yaml concerns me.
Well, using Travis is not mandatory. It will not be able to cover all the testing needs, like running Windows builds using MSVC. Maybe, it will be better to remove Travis builds it at this point.

GamePad64:
Rebased to include new commits.

GamePad64:
Rebased again. People are not enthusiastic about this change =/

noloader:
@Alexander - I was able to test this on a machine with Cmake. Is this expected? Shouldn't Cmake "just work"?

```
$ cd cryptopp
$ cmake
Usage

  cmake [options] <path-to-source>
  cmake [options] <path-to-existing-build>

Specify a source directory to (re-)generate a build system for it in the
current working directory.  Specify an existing build directory to
re-generate its build system.

Run 'cmake --help' for more information.
```

The reason I ask is Make "just works". I can use Posix Make or Gmake, and things start building.

GamePad64:
CMake requires entering source directory, so use `cmake .` (notice the dot). That means, that cmake will search for CMakeLists.txt in current directory.

noloader:
@GamePad64 - At commit 79882d4c7d47d5ef00 I added Cmake artifacts to the GNUMakefile's ***`distclean`*** recipe. If any of the artifacts should be under ***`clean`*** recipe, then please let me know or work-up a pull request.

Issue #80, opened 2015-12-12
[OSX] Incorrect install_name for dynamic library


cawka:
When compiling dynamic version of crypto++ library, it uses `install_name` parameter equal to `libcryptopp.dylib`.  Effectively, if the library is installed, the programs will be able to compile against it, but the at the runtime the library will not be found.

I believe, the default should be `$PREFIX/lib/libcryptopp.dylib`, however there could be other options depending on how the library is used.

(I came across this issue with MacPorts.  It was fixed prior to 5.6.3 update, but now broken again.)

noloader:
Here some background... when we added the additional options on OS X, we did so for versioning requirements as called out at [Dynamic Library Design Guidelines](https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html). Providing ***just*** the library name meant we were honoring a user's choice with a *"regular dependent library"* as described at [Run-Path Dependent Libraries](https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/RunpathDependentLibraries.html).

It sounds like you are asking for the [run-path install names](https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/RunpathDependentLibraries.html). If that's the case, then I think we should take it to the user list for vote. If that's not the case, then can you give us an example of what you would like done (i.e., a diff of the change you are proposing, like using the makefile rule).

cawka:
Thanks for pointers.  I have read them before and reread again.  Unless I have missed it, I didn't see any use case which does not specify absolute name for the install path in some form: real absolute name, with `@executable_path`, `@loader_path`, or `@rpath` (you can check https://wincent.com/wiki/@executable_path,_@load_path_and_@rpath that explains differences).

Which exactly `install_path` is suitable for an application, really depends on the application.  For OSX gui apps and frameworks (that are packaged in .app and .framework folders), `install_path` needs to get properly set for this application, as it depends on where the library is put in the bundle, and there are tools to modify it.   For other apps that just use the shared library installed somewhere, the straightforward way is to specify the absolute path as `install_path`, something like this change:

```patch
diff --git a/GNUmakefile b/GNUmakefile
index d2f5404..1bef87e 100755
--- a/GNUmakefile
+++ b/GNUmakefile
[...]

PR #81, opened 2015-12-12
Makefile improvements

cawka:
First commit was discussed in issue #80 and is part of the mailing list discussion (https://groups.google.com/forum/#!topic/cryptopp-users/_pbRVo7T_OM)

The second commit adds support for DESTDIR, which complements PREFIX variable, but only during the installation phase.  For the current makefile it makes no difference, but can be needed after #80 is merged.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -39,7 +39,7 @@ MACPORTS_COMPILER := $(shell $(CXX) --version 2>&1 | $(EGREP) -i -c "macports")
 
 # Default prefix for make install
 ifeq ($(PREFIX),)
-PREFIX = /usr
+PREFIX = /usr/local
 endif
 
 ifeq ($(CXX),gcc)	# for some reason CXX is gcc on cygwin 1.1.4
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -354,31 +354,31 @@ endif
 
 .PHONY: install
 install:
-	$(MKDIR) -p $(PREFIX)/include/cryptopp $(PREFIX)/lib $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


noloader:
Here's some additional context. It occurred over email.

*****

```
On Mon, Dec 14, 2015 at 12:26 PM, Raphael Straub <raphael@macports.org> wrote:
> Clemens Lang <cal@macports.org> wrote:
>> On Fri, Dec 11, 2015 at 10:55:05PM -0500, Jeffrey Walton wrote:
>>> Sorry to ping you out of cycle.
>>>
>>> Alex discovered a problem with Crypto++'s usage of -install_name. Its
>>> been documented at https://github.com/weidai11/cryptopp/issues/80. A
>>> sister report was filed at https://trac.macports.org/ticket/49987.
>>>
>>> Users will need to build and install the library like so to make
>>> install_name work as desired/expected:
>>>
>>>    make
>>>    make test
>>>    sudo make install PREFIX=/some/path
>>>
>>> My question is, does the proposed fix look compatible with Macport?
>>> Will it cause the Macport project any trouble?
>>>
>>> I don't know how Macport performs the build and install, so I can't
>>> answer the questions.
>>
>> From a short peek that I took: We don't use the make install target at
>> the moment, so that would have to change; it doesn't cause us any
>> trouble, though.
>
> you’re right. Currently the MacPorts Portfile does not use the install target but copies the files manually to a special staging directory called destroot.
>
> I think Crypto++’s makefile could improve their support for package managers like MacPorts by  installing to a staging directory if DESTDIR [1] is set.
>
> Regards
>         Raphael
>
> [1] https://www.gnu.org/prep/standards/html_node/DESTDIR.html
>
```

PR #84, opened 2015-12-14
Microsoft ARM support

m-deckel:
* Adjusted config to compile for Microsoft ARM (e.g. Windows Universal apps)


noloader:
This looks good to me. It looks eerily familiar to a patch I used to have that's been swallowed up by an unresponsive Windows 8 machine...

I'm going to fast track the check-in because someone else was asking about this recently. I need about 4 hours of time to test it on a few different platforms.

PR #88, opened 2015-12-26
Added CRYPTOPP_DATA_DIR option in CMakeLists.txt.

GamePad64:
This isn't really necessary, as this option can be set with `cmake -DCMAKE_CXX_FLAGS="-DCRYPTOPP_DATA_DIR=\"/usr/lib/...\""`
But now you can do `cmake -DCRYPTOPP_DATA_DIR=\"/usr/lib/...\"`, which is simpler.


GamePad64:
Also, provided compatibility with CMake 2.8 (used by Ubuntu LTS, for example)

noloader:
Thank you.

PR #90, opened 2015-12-26
Versioned shared library makefile fixes

tknarr:
Added a makefile target to build the versioned .so if the unversioned name is given as a target.
Added DESTDIR to an ldconf line that was missing it.


noloader:
Thank you sir.

noloader:
@tknarr - Should we make the same changes to [GNUmakefile-cross](https://github.com/weidai11/cryptopp/blob/master/GNUmakefile-cross)?

PR #91, opened 2015-12-26
Apply the same versioned solib changes as to GNUmakefile

tknarr:
Applied the versioned-solib changes to GNUmakefile to GNUmakefile-cross. Removed the DESTDIR references in the process since -cross doesn't use that.


noloader:
Thank you sir.

PR #95, opened 2015-12-29
Fix an issue that caused targets to be rebuilt unnecessarily and prevented updated object files from regenerating targets.

grantae:
The 'public_service' phony target should be an 'order-only' prerequisite so
that it doesn't cause its target to be rebuilt when it is run.
However, object files were specified as order-only while the 'public_service'
target was not. This is reproducible by running 'make' twice. Another effect
is that when these object files change they will not cause the target to be
updated while they are order-only.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -303,6 +303,7 @@ SOLIB_COMPAT_SUFFIX=.$(LIB_MAJOR).$(LIB_MINOR)
 SOLIB_FLAGS=-Wl,-soname,libcryptopp.so$(SOLIB_COMPAT_SUFFIX)
 endif # HAS_SOLIB_VERSION
 
+.PHONY: all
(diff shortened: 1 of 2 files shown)


noloader:
Sounds reasonable. Thank you very much.

In the [near] future, the 5.6.2 warnings about ***`config.h`*** will likely be going away. At that time, I'd like to see the warning on ***`ldconf`*** moved to a 'make install' so the user is not presented with a warning that may not apply.

noloader:
> There's also an issue that libcryptopp.a is rebuilt every time 

Dude, you got it.... Thank you very much. I owe you a cold beer.

By the way, that issue existed long before we added the public service announcement.

grantae:
Glad to contribute. Cryptopp is a great project.

PR #96, opened 2015-12-29
Provided compatibility with CMake 2.8.7

GamePad64:
Okay, finally provided compatibility for CMake 2.8.7, used in Ubuntu 12.04 LTS. Version 2.8.5 was written by mistake, actually.

I had to create "cmake" directory in source tree, as I need to include cmake dynamically, and maintaining multiple cmake files in root dir is not a convenient way to do it.

Also, addressed issue with CMP0042 policy on OS X builds.

There is a little caveat with CMake 2.8.7:
- It doesn't support OBJECT targets, so on CMake 2.8.7 all the source files will be built **twice**, doubling the build time. The first time for the static library, and then for the shared library. CMake 2.8.8 supports this type of target, so it will build each file only once.

----------

This build is tested in an isolated Docker environment with this guest environment:
```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=12.04
DISTRIB_CODENAME=precise
DISTRIB_DESCRIPTION="Ubuntu 12.04.5 LTS"

$ cmake --version     
cmake version 2.8.7

$ g++ --version
g++ (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3
```
Also, tested on Ubuntu 12.04 and OS X at Travis-CI: https://travis-ci.org/GamePad64/cryptopp


noloader:
So what I was thinking here is... Maybe we should provide the Cmake equivalent of a GNU makefile. Maybe it could be called Cmakefile. I'm guessing it would be invoked like:

    cmake -f Cmakefile

I'm only guessing because I have yet to see Cmake actually build the library, which is the whole point of this exercise :) Running Cmake produces a lot of artifacts, but not a `libcryptopp.a` or `cryptest.exe` that a user can actually use.

GamePad64:
CMake is a Makefile generator. First, you run `cmake <source_dir>` it will generate Makefile in working dir (you can open it in text editor and see plain-old `make` syntax), and then `make -f Makefile` (because GNUmakefile overrides Makefile).
I should describe it in wiki.

GamePad64:
Wiki: https://cryptopp.com/wiki/CMake
I will add "Getting CMake" later.

PR #98, opened 2015-12-31
Fix problem, pull request #65 comment 168191115

tknarr:
Fixed the problem with the libcryptopp.so target's right-hand side missing the .so. Also fixed it so we don't have that target when HAVE_SOLIB_VERSION is 0 and the libcryptopp.so$(SOLIB_VERSION_SUFFIX) target will translate to plain libcryptopp.so.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -470,8 +470,10 @@ libcryptopp.a: $(LIBOBJS) | public_service
 	$(AR) $(ARFLAGS) $@ $(LIBOBJS)
 	$(RANLIB) $@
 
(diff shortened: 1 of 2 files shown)


noloader:
Thank you sir. For what its worth, I think these are great additions. The minor pain points are par for the course. I experienced them myself before we put a more disciplined testing process in place. The testing process is working as expected :) Its catching issues proactively, before users experience them.

tknarr:
Thanks. Hopefully this'll be the last of the glitches with this one.

PR #107, opened 2016-01-09
Improve C++Builder support

### LINKED ISSUE #102: Length should be using size_t
Hello, I was having problems compiling Crypto++ with C++Builder XE6 in 64-bit (Clang compiler). Here is the message I was getting:
```
[bcc64 Error] pkcspad.h(79): call to constructor of 'HashIdentifier' (aka 'pair<const byte *, size_t>') is ambiguous
```

My first move was to cast the second parameter as **size_t**:
```cpp
return HashIdentifier(PKCS_DigestDecoration<H>::decoration, (size_t)PKCS_DigestDecoration<H>::length);
```

It was now compiling correctly, but I felt like this was not the proper fix. So instead I changed the **length** property from **unsigned int** to **size_t**:
```cpp
template <class H> class PKCS_DigestDecoration
{
public:
	static const byte decoration[];
	static const size_t length; // It was unsigned int before
};
```

Some other other files were also changed. But I look more accurate to have the result of a **sizeof** in **size_t**.

These are my changes:
```diff
diff --git a/dll.cpp b/dll.cpp
index 72dade9..6328765 100644
--- a/dll.cpp
+++ b/dll.cpp
@@ -24,19 +24,19 @@
 NAMESPACE_BEGIN(CryptoPP)
 
 template<> const byte PKCS_DigestDecoration<SHA1>::decoration[] = {0x30,0x21,0x30,0x09,0x06,0x05,0x2B,0x0E,0x03,0x02,0x1A,0x05,0x00,0x04,0x14};
-template<> const unsigned int PKCS_DigestDecoration<SHA1>::length = sizeof(PKCS_DigestDecoration<SHA1>::decoration);
+template<> const size_t PKCS_DigestDecoration<SHA1>::length = sizeof(PKCS_DigestDecoration<SHA1>::decoration);
 
 template<> const byte PKCS_DigestDecoration<SHA224>::decoration[] = {0x30,0x2d,0x30,0x0d,0x06,0x09,0x60,0x86,0x48,0x01,0x65,0x03,0x04,0x02,0x04,0x05,0x00,0x04,0x1c};
-template<> const unsigned int PKCS_DigestDecoration<SHA224>::length = sizeof(PKCS_DigestDecoration<SHA224>::decoration);
+template<> const size_t PKCS_DigestDecoration<SHA224>::length = sizeof(PKCS_DigestDecoration<SHA224>::decoration);
 
 template<> const byte PKCS_DigestDecoration<SHA256>::decoration[] = {0x30,0x31,0x30,0x0d,0x06,0x09,0x60,0x86,0x48,0x01,0x65,0x03,0x04,0x02,0x01,0x05,0x00,0x04,0x20};
-template<> const unsigned int PKCS_DigestDecoration<SHA256>::length = sizeof(PKCS_DigestDecoration<SHA256>::decoration);
+template<> const size_t PKCS_DigestDecoration<SHA256>::length = sizeof(PKCS_DigestDecoration<SHA256>::decoration);
 
 template<> const byte PKCS_DigestDecoration<SHA384>::decoration[] = {0x30,0x41,0x30,0x0d,0x06,0x09,0x60,0x86,0x48,0x01,0x65,0x03,0x04,0x02,0x02,0x05,0x00,0x04,0x30};
-template<> const unsigned int PKCS_DigestDecoration<SHA384>::length = sizeof(PKCS_DigestDecoration<SHA384>::decoration);
+template<> const size_t PKCS_DigestDecoration<SHA384>::length = sizeof(PKCS_DigestDecoration<SHA384>::decoration);
 
 template<> const byte PKCS_DigestDecoration<SHA512>::decoration[] = {0x30,0x51,0x30,0x0d,0x06,0x09,0x60,0x86,0x48,0x01,0x65,0x03,0x04,0x02,0x03,0x05,0x00,0x04,0x40};
-template<> const unsigned int PKCS_DigestDecoration<SHA512>::length = sizeof(PKCS_DigestDecoration<SHA512>::decoration);
+template<> const size_t PKCS_DigestDecoration<SHA512>::length = sizeof(PKCS_DigestDecoration<SHA512>::decoration);
 
 template<> const byte EMSA2HashId<SHA>::id = 0x33;
 template<> const byte EMSA2HashId<SHA224>::id = 0x38;
diff --git a/pkcspad.cpp b/pkcspad.cpp
index d3d6d2e..a1ed97e 100644
--- a/pkcspad.cpp
+++ b/pkcspad.cpp
@@ -13,16 +13,16 @@ NAMESPACE_BEGIN(CryptoPP)
 
 // more in dll.cpp
 template<> const byte PKCS_DigestDecoration<Weak1::MD2>::decoration[] = {0x30,0x20,0x30,0x0c,0x06,0x08,0x2a,0x86,0x48,0x86,0xf7,0x0d,0x02,0x02,0x05,0x00,0x04,0x10};
-template<> const unsigned int PKCS_DigestDecoration<Weak1::MD2>::length = sizeof(PKCS_DigestDecoration<Weak1::MD2>::decoration);
+template<> const size_t PKCS_DigestDecoration<Weak1::MD2>::length = sizeof(PKCS_DigestDecoration<Weak1::MD2>::decoration);
 
 template<> const byte PKCS_DigestDecoration<Weak1::MD5>::decoration[] = {0x30,0x20,0x30,0x0c,0x06,0x08,0x2a,0x86,0x48,0x86,0xf7,0x0d,0x02,0x05,0x05,0x00,0x04,0x10};
-template<> const unsigned int PKCS_DigestDecoration<Weak1::MD5>::length = sizeof(PKCS_DigestDecoration<Weak1::MD5>::decoration);
+template<> const size_t PKCS_DigestDecoration<Weak1::MD5>::length = sizeof(PKCS_DigestDecoration<Weak1::MD5>::decoration);
 
 template<> const byte PKCS_DigestDecoration<RIPEMD160>::decoration[] = {0x30,0x21,0x30,0x09,0x06,0x05,0x2b,0x24,0x03,0x02,0x01,0x05,0x00,0x04,0x14};
-template<> const unsigned int PKCS_DigestDecoration<RIPEMD160>::length = sizeof(PKCS_DigestDecoration<RIPEMD160>::decoration);
+template<> const size_t PKCS_DigestDecoration<RIPEMD160>::length = sizeof(PKCS_DigestDecoration<RIPEMD160>::decoration);
 
 template<> const byte PKCS_DigestDecoration<Tiger>::decoration[] = {0x30,0x29,0x30,0x0D,0x06,0x09,0x2B,0x06,0x01,0x04,0x01,0xDA,0x47,0x0C,0x02,0x05,0x00,0x04,0x18};
-template<> const unsigned int PKCS_DigestDecoration<Tiger>::length = sizeof(PKCS_DigestDecoration<Tiger>::decoration);
+template<> const size_t PKCS_DigestDecoration<Tiger>::length = sizeof(PKCS_DigestDecoration<Tiger>::decoration);
 
 size_t PKCS_EncryptionPaddingScheme::MaxUnpaddedLength(size_t paddedLength) const
 {
diff --git a/pkcspad.h b/pkcspad.h
index 6a15b49..7532dbf 100644
--- a/pkcspad.h
+++ b/pkcspad.h
@@ -30,7 +30,7 @@ template <class H> class PKCS_DigestDecoration
 {
 public:
 	static const byte decoration[];
-	static const unsigned int length;
+	static const size_t length;
 };
 
 // PKCS_DigestDecoration can be instantiated with the following
```

Is that correct? If so, feel free to patch the code.

Crayon2000:
Here are some fixes required to build the library with Embarcadero C++Builder. This does not address all changes required to allow compilation with bcc32 and bcc64. Other problems might be addressed later.

This pull request fixes #102

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -254,3 +254,18 @@ pip-log.txt
 #Mr Developer
 .mr.developer.cfg
 
+#################
+## C++Builder
+#################
+
+## Ignore C++Builder temporary files and build results.
+## http://docwiki.embarcadero.com/RADStudio/en/File_Extensions_of_Files_Generated_by_RAD_Studio
+
+# Static library file
+*.lib
+
+# User-specific project options
+*.local
+
+# Dependency file
+*.d
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -220,6 +220,9 @@ ifeq ($(findstring asan,$(MAKECMDGOALS)),asan)
 ifeq ($(findstring -fsanitize=address,$(CXXFLAGS)),)
 CXXFLAGS += -fsanitize=address
 endif # CXXFLAGS
+ifeq ($(findstring -fno-omit-frame-pointer,$(CXXFLAGS)),)
+CXXFLAGS += -fno-omit-frame-pointer
+endif # CXXFLAGS
 endif # Asan
 
 # LD gold linker testing. Triggered by 'LD=ld.gold'.
--- a/base32.cpp
+++ b/base32.cpp
@@ -18,7 +18,7 @@ void Base32Encoder::IsolatedInitialize(const NameValuePairs &parameters)
 
 void Base32Decoder::IsolatedInitialize(const NameValuePairs &parameters)
 {
-	BaseN_Decoder::Initialize(CombinedNameValuePairs(
+	BaseN_Decoder::IsolatedInitialize(CombinedNameValuePairs(
 		parameters,
 		MakeParameters(Name::DecodingLookupArray(), GetDefaultDecodingLookupArray(), false)(Name::Log2Base(), 5, true)));
 }
--- a/base64.cpp
+++ b/base64.cpp
@@ -47,6 +47,13 @@ void Base64URLEncoder::IsolatedInitialize(const NameValuePairs &parameters)
 			(Name::Log2Base(), 6, true)));
 }
 
+void Base64Decoder::IsolatedInitialize(const NameValuePairs &parameters)
+{
+	BaseN_Decoder::IsolatedInitialize(CombinedNameValuePairs(
+		parameters,
+		MakeParameters(Name::DecodingLookupArray(), GetDecodingLookupArray(), false)(Name::Log2Base(), 6, true)));
+}
+
 const int *Base64Decoder::GetDecodingLookupArray()
 {
 	static volatile bool s_initialized = false;
@@ -58,6 +65,13 @@ const int *Base64Decoder::GetDecodingLookupArray()
 		s_initialized = true;
 	}
 	return s_array;
+}
+
+void Base64URLDecoder::IsolatedInitialize(const NameValuePairs &parameters)
+{
+	BaseN_Decoder::IsolatedInitialize(CombinedNameValuePairs(
+		parameters,
+		MakeParameters(Name::DecodingLookupArray(), GetDecodingLookupArray(), false)(Name::Log2Base(), 6, true)));
 }
 
 const int *Base64URLDecoder::GetDecodingLookupArray()
--- a/base64.h
+++ b/base64.h
@@ -62,8 +62,7 @@ public:
 	//! \details IsolatedInitialize() is used to initialize or reinitialize an object using a variable
 	//!   number of arbitrarily typed arguments. IsolatedInitialize() does not call Initialize() on
 	//!  attached transformations. If initialization should be propagated, then use the Initialize() function.
-	void IsolatedInitialize(const NameValuePairs &parameters)
-		{CRYPTOPP_UNUSED(parameters);}
+	void IsolatedInitialize(const NameValuePairs &parameters);
     
 private:
 	static const int * CRYPTOPP_API GetDecodingLookupArray();
@@ -123,8 +122,7 @@ public:
 	//! \details IsolatedInitialize() is used to initialize or reinitialize an object using a variable
 	//!   number of arbitrarily typed arguments. IsolatedInitialize() does not call Initialize() on
 	//!  attached transformations. If initialization should be propagated, then use the Initialize() function.
-	void IsolatedInitialize(const NameValuePairs &parameters)
-		{CRYPTOPP_UNUSED(parameters);}
+	void IsolatedInitialize(const NameValuePairs &parameters);
     
 private:
 	static const int * CRYPTOPP_API GetDecodingLookupArray();
--- a/config.h
+++ b/config.h
@@ -449,7 +449,7 @@ NAMESPACE_END
 #endif
 
 // how to declare class constants
-#if (defined(_MSC_VER) && _MSC_VER <= 1300) || defined(__INTEL_COMPILER)
+#if (defined(_MSC_VER) && _MSC_VER <= 1300) || defined(__INTEL_COMPILER) || defined(__BORLANDC__)
 #	define CRYPTOPP_CONSTANT(x) enum {x};
 #else
 #	define CRYPTOPP_CONSTANT(x) static const int x;
--- a/cryptest.sh
+++ b/cryptest.sh
@@ -101,21 +101,15 @@ else
 fi
 
 # Set to 0 if you don't have Asan
-$CXX -x c++ -fsanitize=undefined adhoc.cpp.proto -c -o $TMP/adhoc > /dev/null 2>&1
+$CXX -x c++ -fsanitize=address adhoc.cpp.proto -c -o $TMP/adhoc > /dev/null 2>&1
 if [ "$?" -eq "0" ] && [ "$IS_X86" -ne "0" ]; then
 	HAVE_ASAN=1
 else
 	HAVE_ASAN=0
 fi
 
-# Fixup...
-if [ "$IS_CYGWIN" -ne "0" ] || [ "$IS_MINGW" -ne "0" ]; then
-	HAVE_UBAN=0
-	HAVE_ASAN=0
-fi
-
-# Final fixups for compilers like GCC on ARM64
-if [ "$HAVE_UBSAN" -eq "0" ] || [ "$HAVE_ASAN" -eq "0" ]; then
+# Fixups... Cygwin and MinGW both advertise sanitizer support, but the program fails to link.
+if [ "$HAVE_UBSAN" -eq "0" ] || [ "$HAVE_ASAN" -eq "0" ] || [ "$IS_CYGWIN" -ne "0" ] || [ "$IS_MINGW" -ne "0" ]; then
 	HAVE_UBAN=0
 	HAVE_ASAN=0
 fi
--- a/cryptest.sh
+++ b/cryptest.sh
@@ -38,7 +38,8 @@ IS_LINUX=$(uname -s | grep -i -c linux)
 IS_CYGWIN=$(uname -s | grep -i -c cygwin)
 IS_MINGW=$(uname -s | grep -i -c mingw)
 IS_OPENBSD=$(uname -s | grep -i -c openbsd)
-IS_INTEL=$(uname -m | egrep -i -c "(i386|i586|i686|amd64|x86_64)")
+IS_X86=$(uname -m | egrep -i -c "(i386|i586|i686|amd64|x86_64)")
+IS_X64=$(uname -m | egrep -i -c "(amd64|x86_64)")
 IS_PPC=$(uname -m | egrep -i -c "(Power|PPC)")
 
 # We need to use the C++ compiler to determine if c++11 is available. Otherwise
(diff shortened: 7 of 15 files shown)


noloader:
@Crayon2000 - You titled the PR *"reduce [compile] errors"*. If you don't mind me asking:

* what are all of the error?
* why are only some of them cleared?

noloader:
Discussion at [C++ Builder pull request](http://groups.google.com/d/msg/cryptopp-users/kKiFCN-URts/RDnR3u6NBAAJ)

Crayon2000:
> * what are all of the error?
> * why are only some of them cleared?

@noloader, there are not a lot of compiler errors left in the code.

#### Here are two errors related to the use of MACRO

`[bcc32 Error] seckey.h(83): E2313 Constant expression required`
line 83 is:
```cpp
#if (M==INT_MAX)	// Coverity and result_independent_of_operands
```
`[bcc32 Error] seckey.h(194): E2313 Constant expression required`
line 194 is:
```cpp
#if MIN_KEYLENGTH > 0
```

The reason why I did not fixed them is that I have no idea what is causing them. I'm guessing that the compiler does not like this:
```cpp
	//! \brief The minimum key length used by the algorithm provided as a constant
	//! \details MIN_KEYLENGTH is provided in bytes, not bits
	CRYPTOPP_CONSTANT(MIN_KEYLENGTH=N)
```

#### Project
Also the project included in the repository is old. I think it is C++Builder 2006. At that time Win64 was not supported and Unicode neither. The second feature might be an issue depending on how the project will be updated to your current version of C++Builder. Sometimes by default it is set to **char** and other times to **wchar_t**.

In that case you will get those errors:
```
[bcc32 Error] fipstest.cpp(311): E2342 Type mismatch in parameter 'lpOutputString' (wanted 'const wchar_t *', got 'const char *')
[bcc32 Error] fipstest.cpp(405): E2342 Type mismatch in parameter 'lpOutputString' (wanted 'const wchar_t *', got 'const char *')
[bcc32 Error] fipstest.cpp(424): E2342 Type mismatch in parameter 'lpOutputString' (wanted 'const wchar_t *', got 'const char *')
```
Those 3 errors could be fixed by changing `OutputDebugString` to `OutputDebugStringA` to force ANSI.

But the proper fix, *I think*, would be to changed the option **_TCHAR maps to** to **char**
The problem is that I don't have an older version of the IDE installed on my PC. I only have XE6, XE7 and XE8. So that's why I didn't change the project.

#### Summary
I have decided to only change things that I was sure would not have any impact on other compilers or on other C++Builder versions. I though it was better to have those rather than nothing.

Tell me if I did the good choices.

noloader:
> `#if (M==INT_MAX)    // Coverity and result_independent_of_operands`
> ...
> `#if MIN_KEYLENGTH > 0`

OK, so the first one was due to a Coverity finding. I don't recall about the second one. It appears Borland's compiler is about as lame as Clang's when it comes to identifying a `constexpr`. When we do something like below during compilation:

    CRYPTOPP_CONSTANT(MIN_KEYLENGTH=16)

`MIN_KEYLENGTH=16` ***never*** changes in this universe under the laws of nature as we know them. But Clang and Borland can't seem to figure that out...

To work around the issue, try this:

    #if !defined(__BORLAND__) && (M==INT_MAX)
    ...
    #endif

That *should* short circuit the evaluation of the expression so the `(M==INT_MAX)` is *not* evaluated.

If that works for you, then let me know and I'll test it against Coverity to ensure its OK. We need to throttle Coverity scans because they only give us a limited number of them in a week or a month.

noloader:
> Sometimes by default it is set to char and other times to `wchar_t`...

OK, we encountered this under Visual Studio too (IIRC). For Visual Studio, we changed all the character sets in all the project files to ***`MBCS`***. We only had to do it for one or two projects.

> `[bcc32 Error] fipstest.cpp(311): E2342 Type mismatch in parameter 'lpOutputString' (wanted 'const wchar_t *', got 'const char *')`

Has it always been that way? Or is this from the x64 project you added? Or is it a byproduct of upgrading the Borland project files?

Is it possible for you to change everything to use a multi-byte character set (MBCS) rather than UTF16 (Unicode)?

I tried to find a used copy of C++ Builder on Amazon, but most of the results are *"currently unavailable"*. I found one seller with [C++ Builder 2006](http://www.amazon.com/dp/B000IAJZAI/), but its a bit pricey $400 USD (almost 1.5x the cost of Visual Studio 2015 Pro). I wrote to the seller and asked to the reduce the price. I hope he is willing to help us.

I'm not sure what else we can do. I don't think I can really help with this one.

noloader:
> I have decided to only change things that I was sure would not have any impact on other compilers or on other C++Builder versions. I though it was better to have those rather than nothing.
>
> Tell me if I did the good choices.

Yeah, that's fine. There's no right or wrong; and no good or bad. We are trying to gauge the depth and breadth of issues under C++ Builder and its compilers.

noloader:
From the pull request files... Lets remove this one:

    +# Compiled resource file
    +*.res

The problem, is, that's not an output artifact under some IDEs. That is, `*.res` shows up as a real file meant to be version controlled. I think either Visual Studio or Eclipse uses them (but I'm not current on a lot of the project files used by many of the IDEs).

Crayon2000:
#### Macro
First of all, the problem with the MACRO only occurs with bcc32, the old Win32 compiler. In case you didn't know C++Builder as a bunch of different compilers: http://docwiki.embarcadero.com/RADStudio/en/Command-Line_Interfaces_of_C%2B%2B_Compilers

I have tried this:
```cpp
#if !defined(__BORLANDC__) && MIN_KEYLENGTH > 0
		if (keylength < (size_t)MIN_KEYLENGTH)
			return MIN_KEYLENGTH;
		else
#endif
```
Unfortunately the second condition is evaluated when it's on the same line!

So instead I need to do something like this:
```cpp
#if !defined(__BORLANDC__)
#if MIN_KEYLENGTH > 0
		if (keylength < (size_t)MIN_KEYLENGTH)
			return MIN_KEYLENGTH;
		else
#endif
#endif
```
And 
```cpp
#if !defined(__BORLANDC__)
#if (M==INT_MAX)	// Coverity and result_independent_of_operands
		if (rounds < MIN_ROUNDS)
			throw InvalidRounds(alg ? alg->AlgorithmName() : "VariableRounds", rounds);
#else
		if (rounds < MIN_ROUNDS || rounds > MAX_ROUNDS)
			throw InvalidRounds(alg ? alg->AlgorithmName() : "VariableRounds", rounds);
#endif
#endif
```
What do you think, should I commit that? In the second case, validation will be removed completely.
An other possibility is this:
```cpp
#if defined(__BORLANDC__)
		if (rounds < MIN_ROUNDS || rounds > MAX_ROUNDS)
			throw InvalidRounds(alg ? alg->AlgorithmName() : std::string("VariableRounds"), rounds);
#elif (M==INT_MAX)	// Coverity and result_independent_of_operands
		if (rounds < MIN_ROUNDS)
			throw InvalidRounds(alg ? alg->AlgorithmName() : "VariableRounds", rounds);
#else
		if (rounds < MIN_ROUNDS || rounds > MAX_ROUNDS)
			throw InvalidRounds(alg ? alg->AlgorithmName() : "VariableRounds", rounds);
#endif
```

#### Project: MBCS vs UTF16
C++Builder 2009 introduced Unicode. From what I see Crypto++ should support C++Builder 2010 and up. The problem is the current project file was created with C++Builder 2006. So, it can not be easily changed manually by editing the project file content. Normally, that's the section that is required in the *.cbproj* file:
```xml
    <PropertyGroup Condition="'$(Base)'!=''">
        <_TCHARMapping>char</_TCHARMapping>
    </PropertyGroup>
```

> I'm not sure what else we can do. I don't think I can really help with this one.

C++Builder 2010 was once installed at work. If I still have it on my PC I will update those files and set MBCS as default:
* cryptest_bds.bdsgroup
* cryptlib_bds.bdsproj
* cryptest_bds.bdsproj

#### Res file
The *.res was removed from .gitignore in my last commit.

noloader:
Yeah, this looks like the one:

```
#if defined(__BORLANDC__)
        if (rounds < MIN_ROUNDS || rounds > MAX_ROUNDS)
            throw InvalidRounds(alg ? alg->AlgorithmName() : std::string("VariableRounds"), rounds);
#elif (M==INT_MAX)  // Coverity and result_independent_of_operands
        if (rounds < MIN_ROUNDS)
            throw InvalidRounds(alg ? alg->AlgorithmName() : "VariableRounds", rounds);
#else
        if (rounds < MIN_ROUNDS || rounds > MAX_ROUNDS)
            throw InvalidRounds(alg ? alg->AlgorithmName() : "VariableRounds", rounds);
#endif
```

noloader:
Thank you very much.

Sorry for the drawn out process. It goes much faster when I can test it locally.

Crayon2000:
@noloader it was a pleasure to help you and the Crypto++ project.

PR #110, opened 2016-01-12
Fixes a compiler error with C++Builder XE6

Crayon2000:
This is the generated error with CLANG 64-bit:
[bcc64 Error] pkcspad.h(79): call to constructor of 'HashIdentifier' (aka 'pair<const byte *, size_t>') is ambiguous

Casting the second parameter as size_t will fix the problem. This fix is not required in C++Builder XE7 and XE8, it must have been fixed.

### DIFF
--- a/pkcspad.h
+++ b/pkcspad.h
@@ -76,7 +76,7 @@ public:
 		{
 			static HashIdentifier Lookup()
 			{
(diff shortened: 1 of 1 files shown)


noloader:
Thank you very much.

PR #111, opened 2016-01-12
With C++Builder we need to include stdlib.h to use wcstombs

Crayon2000:
wcstombs was supported in C++Builder 2010 : http://docwiki.embarcadero.com/RADStudio/2010/en/Wcstombs

### DIFF
--- a/misc.h
+++ b/misc.h
@@ -1,4 +1,4 @@
-﻿// misc.h - written and placed in the public domain by Wei Dai
+// misc.h - written and placed in the public domain by Wei Dai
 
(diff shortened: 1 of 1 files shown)


noloader:
@Crayon2000 - I see there's a revert. Please forgive my ignorance... Is this ready to merge?

Crayon2000:
@noloader yes it's ready. My editor changed the format from UTF8 to UTF8 BOM in my first commit. I just undid that change in the second one.

noloader:
Thank you very much.

PR #114, opened 2016-01-13
Integer Patch

DevJPM:
This is the Integer Patch as requested on the mailing list (i.e. includes the LE code and the std::showbase stuff)

Sorry for the change in OID, I copied the branch from my main fork which already had the ECPM changes and screwed up a little there, but the change is completely minor (adding tabs).


noloader:
This looks good.

Would it be possible for you to make the case of [`order == BIG_ENDIAN`](https://github.com/weidai11/cryptopp/pull/114/files#diff-979626659f96981c10c3c77d608c12f3R3150) the first case? Its the dominant case, and it will help things run a little faster.

noloader:
> Sorry for the change in OID

This should not be a problem. We should be able to [merge this and ignore the whitespace chages](http://stackoverflow.com/q/9776527).

DevJPM:
> Would it be possible for you to make the case of `order == BIG_ENDIAN` the first case?

@noloader done.

noloader:
Done. Thank you very much.

PR #116, opened 2016-01-15
Made MeterFilter Get*() functions const

greenius:
Made the MeterFilter Get functions const as they return information about the filter without any side effects and allow client code to be const correct.


noloader:
Thank you very much.

PR #142, opened 2016-03-03
Remove unwanted assert

Giga-gg:
When decoding a key, Crypto++ may throw an exception (if there is a format error).
I want to be able to catch exception, but it is not possible in debug mode. Here how it goes:

* An exception is thrown during the decoding process
* Before it can be caught, the ```BERGeneralDecoder``` destructor is called.
* The destructor calls ```MessageEnd()``` which in turn throw an exception
* As throwing exception in destructor is bad, the exception get caught and an ```assert(0)``` is called

So in debug mode, each time there is a decoding error, the ```assert(0)``` gets called, and the program stops.

To fix this I just removed the ```assert(0)``` in the catch block.
(This was tested with visual studio 2015)

### DIFF
--- a/asn.cpp
+++ b/asn.cpp
@@ -405,14 +405,14 @@ void BERGeneralDecoder::Init(byte asnTag)
 
 BERGeneralDecoder::~BERGeneralDecoder()
 {
-	try	// avoid throwing in constructor
+	try	// avoid throwing in destructor
 	{
 		if (!m_finished)
 			MessageEnd();
 	}
 	catch (const Exception&)
 	{
-		assert(0);
+		// assert(0);
 	}
 }
 


noloader:
I don't really want to remove the assert, but we will should since its crashing your program.

Posix's assert behavior is totally useless. Asserting is a debugging and diagnostic feature, and asserting creates self debugging programs. What on earth motivated Posix to crash a program when its under a debugger?

We have a fix for the useless Posix behavior... Its `CRYPTOPP_ASSERT` that calls `SIGTRAP` so it snaps the debugger rather than killing the program. You can see it at [trap.h](https://github.com/weidai11/cryptopp/blob/master/trap.h).

We are waiting to cut-in `CRYPTOPP_ASSERT`. It will probably occur at a minor version bump, like 5.6.x → 5.7. We did not want to do it during a revision bump because some folks use assert in release/production to force a crash.

We think using an assert to force a crash in production is another abuse of assert since an assert is a debugging and diagnostic feature, but that's a different complaint.... Folks who want it can define `CRYPTOPP_POSIX_ASSERT` to retain Posix assert behavior.

PR #150, opened 2016-03-24
Add default OS X NDK root

lewellyn:
The standard NDK installed by Android Studio on OS X lives at $HOME/Library/Android/sdk/ndk-bundle. As this is a well-known location that seems unlikely to change, it makes sense to auto-detect it.


noloader:
Thank you very much.

PR #151, opened 2016-03-29
Fixed linker error when using VS2015 /MD

bretthall:
Without this change links under Visual Studio 2015 using the `/MDd` switch ("multithreaded debug dll" c-runtime) fail with a `LNK2005: __crt_debugger_hook already defined in msvcrtd.lib(utility_desktop.obj)`
error.

### DIFF
--- a/fipstest.cpp
+++ b/fipstest.cpp
@@ -21,7 +21,11 @@
 #else
 #define _CRT_DEBUGGER_HOOK __crt_debugger_hook
 #endif
+#if _MSC_VER < 1900
 extern "C" {_CRTIMP void __cdecl _CRT_DEBUGGER_HOOK(int);}
+#else
+extern "C" {void __cdecl _CRT_DEBUGGER_HOOK(int); }
+#endif
 #endif
 #endif
 


noloader:
Thanks @bretthall. How can this be tested? What did you need to do?

The reason I ask is I don't have VS2015 (I'm at VS2013), so I need to ask some folks on the User Group to test it and report back.

bretthall:
We got the link error when building our application that statically links with Crypto++. But in trying to solve the problem (originally thought it was something in our code) I tried building Crypto++ as a DLL and encountered the linker problem there as well. So just building Crypto++ as a DLL using VS2015 should be enough to test the patch (if it builds in debug mode then the patch works).

I'm out of town at the moment and don't have access to the machine at work that has VS2015 on it. I'll be back on Monday so I can get you a more detailed recipe then.

bretthall:
OK, here's more detailed instructions:

- Clone the git repository
- Copy the solution and project files out of VS2010.zip into the top-level of the source tree
- Open cryptest.sln in Visual Studio 2015
- Open the properties of the cryptdll project
- Under "C/C++, Code Generation" change "Runtime Library" to be "Multi-threaded Debug DLL (/MDd)"
- repeat this for the cryptest and cryptlib projects
- Build the cryptdll project in Debug mode
- Without the fix you will get "ucrtd.lib(ucrtbased.dll) : error LNK2005: __crt_debugger_hook already defined in MSVCRTD.lib(utility_desktop.obj)" errors in the cryptest and cryptdll projects. With the patch everything should build fine.

zabulus:
I can confirm that this patch fixes the issue

noloader:
> I tried building Crypto++ as a DLL and encountered the linker problem there as well.

Also worth mentioning... [FIPS DLL](http://www.cryptopp.com/wiki/FIPS_DLL) on the Crypto++ wiki. The DLL is *not* a general purpose Windows DLL. Its special purpose, and only provided FIPS approved algorithms.

When you use the FIPS DLL, you still need to build and link to the static library because the static library provides support and other goodies, like `HexEncoder`, `FileSource`, `FileSink` etc.

bretthall:
Yeah, but even if you don't use the DLL you might encounter the linker error in projects that link to the static library. We have two projects that use Crypto++, one of them hit the linker error and one didn't. We have no idea what the difference is that caused one to get the error and the other not to. But the patch fixed the linker error in the project that saw it.

bretthall:
Oops, hit the close button on accident when I just meant to comment.

noloader:
@bretthall - related, do you have Windows scripting and msbuild skills?

We have a "test script from hell" for Unix and Linux at [`cryptest.sh`](http://github.com/weidai11/cryptopp/blob/master/cryptest.sh). Here are the head notes for the script:

> \# This is a test script that can be used on some Linux/Unix/Apple machines to automate building the
> \# library and running the self test with various combinations of flags, options, and conditions.

We lack one for Windows (`cryptest.cmd`?). I could use one for Windows for testing changes like this.

If you have the skills, then could you build out a basic script? If so, then I'll open a bug report.

bretthall:
Unfortunately, I don't do much with windows batch files or MSBuild. I usually switch to cygwin bash for anything complicated when it comes to scripting. For building on windows its either Visual Studio or CMake.

noloader:
@bretthall,

Sorry for the delay in looping back to this...

One final question. According to the MSDN docs on [/MD, /MT, /LD (Use Run-Time Library)](https://msdn.microsoft.com/en-us/library/2kzt1wy3.aspx):

* `/MT` causes macro definitions for `_MT` (Crypto++ uses this)
* `/MD` causes macro definitions for `_MT` and `_DLL` (Crypto++ does not use this)

So my question is, can we guard this block with `if defined(_MT) && !defined(_DLL)` to avoid it cross-pollinating or spilling out into other parts of a user program?

zabulus:
If the change fixes the issue, I am ok with propose

bretthall:
If we guard the block with `if defined(_MT) && !defined(_DLL)` won't that disable the fix when using `/MD` which is the setting that requires the fix? I didn't test it, but there's also a chance that it the same problem would occur when using `/MT` and statically linking the library. So guarding on the `_DLL` macro probably isn't a good idea.

noloader:
Thank you very much.

PR #152, opened 2016-04-05
Fix Visual Studio compilation with _HAS_EXCEPTIONS=0

bozaro:
Bug information: https://connect.microsoft.com/VisualStudio/feedback/details/1600701/type-info-does-not-compile-with-has-exceptions-0

For example:
```
C:\Work\github\cryptopp>cl /c /Tp cryptlib.h /Focryptlib.h.obj /D_HAS_EXCEPTIONS=0
Microsoft (R) C/C++ Optimizing Compiler Version 18.00.31101 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

cryptlib.h
c:\work\github\cryptopp\cryptlib.h(292) : error C2039: 'type_info' : is not a member of 'std'
c:\work\github\cryptopp\cryptlib.h(298) : error C2039: 'type_info' : is not a member of 'std'
c:\work\github\cryptopp\cryptlib.h(302) : error C2039: 'type_info' : is not a member of 'std'
c:\work\github\cryptopp\cryptlib.h(305) : error C2039: 'type_info' : is not a member of 'std'
c:\work\github\cryptopp\cryptlib.h(306) : error C2039: 'type_info' : is not a member of 'std'
c:\work\github\cryptopp\cryptlib.h(392) : error C2039: 'type_info' : is not a member of 'std'
c:\work\github\cryptopp\cryptlib.h(437) : error C2039: 'type_info' : is not a member of 'std'
```

### DIFF
--- a/stdcpp.h
+++ b/stdcpp.h
@@ -6,13 +6,6 @@
 #include <intrin.h>
 #endif
 
-#ifdef _MSC_VER
-// Workaround for: https://connect.microsoft.com/VisualStudio/feedback/details/1600701/type-info-does-not-compile-with-has-exceptions-0
-namespace std {
-  using ::type_info;
-}
-#endif
-
 #include <string>
 #include <memory>
 #include <exception>
@@ -26,6 +19,13 @@ namespace std {
 #include <list>
 #include <map>
 #include <new>
+
+#if defined(_MSC_VER) && (_MSC_VER < 1900) && defined(_HAS_EXCEPTIONS) && (_HAS_EXCEPTIONS == 0)
+// Workaround for: https://connect.microsoft.com/VisualStudio/feedback/details/1600701/type-info-does-not-compile-with-has-exceptions-0
+namespace std {
+  using ::type_info;
+}
+#endif
 
 #if _MSC_VER >= 1600
 // for make_unchecked_array_iterator
--- a/stdcpp.h
+++ b/stdcpp.h
@@ -6,6 +6,13 @@
 #include <intrin.h>
 #endif
 
+#ifdef _MSC_VER
+// Workaround for: https://connect.microsoft.com/VisualStudio/feedback/details/1600701/type-info-does-not-compile-with-has-exceptions-0
+namespace std {
+  using ::type_info;
+}
+#endif
+
 #include <string>
 #include <memory>
 #include <exception>


noloader:
Thanks Bozaro. This one is going to need some testing to ensure we can safely make the change.

Can you confirm it applies to Visual Studio 2015 and Visual Studio 2015 SP1? I should be able to test Visual Studio 2002 - Visual Studio 2013 (I have them in my test environment).

bozaro:
Unfortunately, I use Visual Studio 2013 :(

noloader:
Thanks again Bozaro. I'm kind of struggling with this at higher levels.

As far as I know, there's no way to disgorge C++ from exceptions in Crypto++. Crypto++ relies on them to convey errors. In fact, I could not use Crypto++ in eVC 3.0 or 4.0 projects because Microsoft's early embedded C++ compiler did not provide exception support. We had to wait for eVC 4.0 or 5.0 (IIRC).

Exception support is also listed as one of the four requirements at [External Dependencies](http://cryptopp.com/wiki/External_Dependencies). trying to use Crypto++ without exceptions is like trying to use it without namespaces (but probably more painful).

According to [STL uses exception handling even if _HAS_EXCEPTIONS is 0](http://connect.microsoft.com/VisualStudio/feedback/details/1015473), Microsoft considers  `_HAS_EXCEPTIONS=0` undocumented and unsupported. It seems to be from early STL libraries like Dinkumware and STLport. As far as I know, those STL libraries have provided full exception support for the last 15 years or so.

[Speak up if you use #define _HAS_EXCEPTIONS](https://blogs.msdn.microsoft.com/nikolad/2005/06/29/speak-up-if-you-use-define-_has_exceptions/) from 2005 appears to be a call for use cases relying on `_HAS_EXCEPTIONS=0`. It sounds like it was the death knell for `_HAS_EXCEPTIONS`.

I'm not sure there's much we can do here.

noloader:
Thanks again Bozaro. I don't want to dismiss this report too soon. I asked for some feedback on the mailing list at [Pull Request 152: Fix Visual Studio compilation with _HAS_EXCEPTIONS=0](https://groups.google.com/d/msg/cryptopp-users/LLxMmGTk_hM/xxPobmewCAAJ).

bozaro:
Shipping game build on latest Unreal Engine 4.11 builded without std::exception by default. CryptoPP is used in some optional Unreal Engine modules.

I try to use this module and get compilation issue. My compilation error looks like a Microsoft C++ library implementation issue.

MarcelRaad:
With Visual Studio 2015's default toolset, this is not needed, but compiles fine.

With Visual Studio 2015's clang/c2 toolset, this leads to the error:
>1>./stdcpp.h(11,9): error : no member named 'type_info' in the global namespace

So maybe this could be restricted to
`#if defined(_MSC_VER) && (_MSC_VER < 1900) && defined(_HAS_EXCEPTIONS) && (_HAS_EXCEPTIONS == 0)`
?

bozaro:
@MarcelRaad 

I add this change and move code sniplet after `#include <typeinfo>`.

noloader:
Thank you very much.

Issue #99, opened 2016-01-01
MSVC2015 (win64) CMakeLists.txt issues


ZahlGraf:
Hi

first of all: Thanks for this great library, it helped me a lot in a current student project!

However there were some compile issues with msvc2015 and CMake:
- for MSVC some assembler sources must be add
- also Assembler must be enabled and correct flags must be set
- there was also another source winpipes.cpp missing

Furthermore the define CRYPTO_DATA_DIR needs quote symbols (for all compilers and systems)

I created a small patch, that works for me. Maybe this also helps other people out there.
I tested the patch for Win64 build (MSVC2015), Linux64 (GCC) build and Android (Android NDK r10e) build. In all builds the debug and release version works well. I did not test the shared libraries, since they are limited anyway. 
However currently I have no time to test Win32 builds or other compilers and systems like MacOSX and CLang. That's why I did not create a pull request.

This is the patch, have a lot of fun with it :)

```patch
From 1531ab255d84b5c774284031927293f18f7c0254 Mon Sep 17 00:00:00 2001
From: Andre Netzeband <there_is_no@mail.de>
Date: Fri, 1 Jan 2016 18:32:34 +0100
Subject: [PATCH] Corrected missing quote symbols for crypto-data-dir
 definition. Added special settings for MSVC: Enable Assembler, set assembler
 flags for 32bit and 64bit added missing assembler files and source files.

---
 CMakeLists.txt | 17 ++++++++++++++++-
 1 file changed, 16 insertions(+), 1 deletion(-)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 7ca9aea..8fa8631 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,13 +48,15 @@ if(DISABLE_AESNI)
 	add_definitions(-DCRYPTOPP_DISABLE_AESNI)
 endif()
 if(NOT CRYPTOPP_DATA_DIR STREQUAL "")
-	add_definitions(-DCRYPTOPP_DATA_DIR=${CRYPTOPP_DATA_DIR})
+	add_definitions(-DCRYPTOPP_DATA_DIR="${CRYPTOPP_DATA_DIR}")
 endif()
 
 #============================================================================
 # Sources & headers
 #============================================================================
 
+math(EXPR BITS "8*${CMAKE_SIZEOF_VOID_P}")
+
 # Library headers
 file(GLOB cryptopp_HEADERS *.h)
 
[...]

PR #156, opened 2016-04-07
Fix MSVC2015 (win64) CMakeLists.txt issues #99

zabulus:
Patch based on commit https://github.com/GamePad64/cryptopp/commit/a2d3ac19b14bdc8c51ab4991b9eeeef7883bfec8
I've tested the original patch, but it doesn't work because of cmake bug: https://cmake.org/Bug/view.php?id=14711.
Used workaround ```set_source_files_properties``` from the link above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ set(cryptopp_VERSION_PATCH 3)
 include(GNUInstallDirs)
 include(TestBigEndian)
 include(CheckCXXSymbolExists)
+
 #============================================================================
 # Settable options
 #============================================================================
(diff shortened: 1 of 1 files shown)


noloader:
Does anyone have any comments on this? If there are no objections, then I'll merge it.

CMakeLists.txt
```diff
@@ -78,10 +77,26 @@ set(cryptopp_SOURCES
...
+ 	if(CMAKE_CL_64) 		
+ 		list(APPEND cryptopp_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/x64dll.asm)
+ 		list(APPEND cryptopp_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/x64masm.asm)
+        set_source_files_properties(${CMAKE_CURRENT_SOURCE_DIR}/x64dll.asm PROPERTIES COMPILE_FLAGS "/D_M_X64")
+        set_source_files_properties(${CMAKE_CURRENT_SOURCE_DIR}/x64masm.asm PROPERTIES COMPILE_FLAGS "/D_M_X64")
+        set_source_files_properties(${CMAKE_CURRENT_SOURCE_DIR}/rdrand.asm PROPERTIES COMPILE_FLAGS "/D_M_X64")
+ 	else()    
+        set_source_files_properties(${CMAKE_CURRENT_SOURCE_DIR}/x64dll.asm PROPERTIES COMPILE_FLAGS "/D_M_X86")
+        set_source_files_properties(${CMAKE_CURRENT_SOURCE_DIR}/x64masm.asm PROPERTIES COMPILE_FLAGS "/D_M_X86")
+        set_source_files_properties(${CMAKE_CURRENT_SOURCE_DIR}/rdrand.asm PROPERTIES COMPILE_FLAGS "/D_M_X86")
```
noloader: The pickup of `/D_M_X86` and `/D_M_X64` was good. However, there's a subtle issue remaining: both `rdrand.cpp` and `rdrand.asm` produce an outfile of of `rdrand.obj`. Here's how we handle it in the [test nmake makefile](http://github.com/weidai11/cryptopp/blob/master/cryptest.nmake#L119):

* `rdrand.cpp` produces output artifact `rdrand.obj`
* `rdrand.asm` with `/D_M_X86` produce output artifact `rdrand-x86.obj`
* `rdrand.asm` with `/D_M_X64` produce output artifact `rdrand-x64.obj`

Then, link with the pair (`rdrand.obj`, `rdrand-x86.obj`) or (`rdrand.obj`, `rdrand-x64.obj`)

noloader:
Thank you very much.

PR #157, opened 2016-04-21
Adjusted ChaCha for dynamic round selection

DevJPM:
I've adjusted the ChaCha code to allow selection of the round number at
run-time and to improve readibility (slightly), this patch also makes ChaCha more consistent with Salsa

### DIFF
--- a/chacha.cpp
+++ b/chacha.cpp
@@ -27,8 +27,7 @@ void ChaCha_TestInstantiations()
 }
 #endif
 
-template <unsigned int R>
-void ChaCha_Policy<R>::CipherSetKey(const NameValuePairs &params, const byte *key, size_t length)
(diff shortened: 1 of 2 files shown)


noloader:
Related, see [Salsa20_Policy::CipherSetKey](http://github.com/weidai11/cryptopp/blob/master/salsa.cpp#L42).

DevJPM:
@noloader I changed it (the `CipherSetKey()` function) locally but changes somehow didn't survive the push to GitHub. I've now adjusted the PR.

noloader:
Round selection occurs through the template parameter and enjoys a small speedup. I'd prefer to avoid the dynamic round selection.

The `typedefs` look OK, so they should be able to be changed. The PR has conflicts, so I can't merge it at the moment. I'm going to close it for the time being.

Maybe you could re-issue the PR for the `typederf` changes?

PR #171, opened 2016-05-08
Fix _WIN32_WINNT for Windows Store Apps >= 10.*

zabulus:
If use CMake for WindowsStore 10.0, _WIN32_WINNT not set and it compiles library without CNG API


noloader:
Thank you very much.

PR #172, opened 2016-05-10
Check for SSE4 support before using SSE4.1 instruction

MarcelRaad:
In a 32-bit Windows program compiled with Visual C++ 2013 Update 5, we
sometimes get crashes because of an exception "0xC000001D: Illegal Instruction"
on the pextrd (_mm_extract_epi32) instruction. Explicitly check for SSE4
support instead of only AES-NI before using this SSE4.1 instruction.

### DIFF
--- a/rijndael.cpp
+++ b/rijndael.cpp
@@ -217,9 +217,9 @@ void Rijndael::Base::UncheckedSetKey(const byte *userKey, unsigned int keylen, c
 
 	word32 *rk = m_key;
 
-#if (CRYPTOPP_BOOL_AESNI_INTRINSICS_AVAILABLE && (!defined(_MSC_VER) || _MSC_VER >= 1600 || CRYPTOPP_BOOL_X86 || CRYPTOPP_BOOL_X32))
+#if (CRYPTOPP_BOOL_AESNI_INTRINSICS_AVAILABLE && CRYPTOPP_BOOL_SSE4_INTRINSICS_AVAILABLE && (!defined(_MSC_VER) || _MSC_VER >= 1600 || CRYPTOPP_BOOL_X86 || CRYPTOPP_BOOL_X32))
 	// MSVC 2008 SP1 generates bad code for _mm_extract_epi32() when compiling for X64
-	if (HasAESNI())
+	if (HasAESNI() && HasSSE4())
 	{
 		static const word32 rcLE[] = {
 			0x01, 0x02, 0x04, 0x08,


noloader:
Thanks Marcel. I'm kind of aware of this pain point even though I have not given it attention.

What hardware are you using to reproduce it?

MarcelRaad:
Unfortunately I haven't been able to reproduce this myself, I received the crash dumps via Windows Error Reporting. All the crashing machines were different HP and Fujitsu notebooks. I have requested additional CPU information via WER, but haven't received it yet.

noloader:
OK, thanks. If you get word, then please let me know so I can set up a test case for it. My testing infrastructure includes Libvirt/Qemu, so I can call out processor features on demand.

noloader:
@MarcelRaad - I added some tests to the test script that may catch this. See [Commit cdd88236afb2, Add test for modern compiler and old hardware](http://github.com/weidai11/cryptopp/commit/cdd88236afb2c4726f15c5fadc3ae84b585cc7fa).

I think we should probably perform a disassembly of some object files to ensure the instructions are generated and survive optimizations.

noloader:
@MarcelRaad - I also noticed some mildly older hardware, like a Core2 Duo, will have an interesting mix of features. I imagine there are other interesting mixes. If you are building Crypto++ offline and then distributing as part of your product, then you may want to:

```
export CXXFLAGS="-g2 -O2 -match=native -maes -mrdrnd -mrdseed -mno-sse4"
```

You should be safe to enable AES and friends unconditionally because [they are compartmentalized and guarded](http://github.com/weidai11/cryptopp/blob/master/cpu.h#L152). SSE3, SSSE3 and SSE4 are different stories.

Another wrinkle is Clang managed to ship a compiler with SSE2, SSE3, SSE4 but not SSSE3. You may or may not encounter it. Also see [LLVM Issue 20213](http://llvm.org/bugs/show_bug.cgi?id=20213).

noloader:
@MarcelRaad,

> I think we should probably perform a disassembly of some object files to ensure the instructions are generated and survive optimizations.

A quick follow up... I've added some tests to [`cryptest.sh`](https://github.com/weidai11/cryptopp/blob/master/cryptest.sh#L1058) which perform the disassembly to ensure we are mostly getting the code generation we expect.

Unfortunately, the testing only occurs under Linux, OS X and Solaris. i don't have anything for Windows at the moment.

MarcelRaad:
Thanks a lot! Sorry for not responding to your previous comments, I thought I could get back soon with more info on the hardware, but it seems like noone affected is using our older crashing version anymore :-(

noloader:
@MarcelRaad,

I found one of those processors... An [Intel Atom D2550](https://ark.intel.com/products/65470/Intel-Atom-Processor-D2550-1M-Cache-1_86-GHz) has SSE2, SSE3, SSSE3 and AES. It lacks SSE4 and other goodies.

noloader:
@MarcelRaad,

I was thinking... We should probably provide an alternate code path for the SSE4 instructions in the AES-NI code path. A low-end Atom needs all the help it can get. Taking away AES-NI due to lack of SSE4 is probably the wrong the choice in the bigger picture. We might also see a similar circumstance in SHA. The SHA code path uses one or two SSE4 instructions.

PR #174, opened 2016-05-31
Fix clang warning

MarcelRaad:
In 64-bit mode, clang 3.8 warns:
> cast to 'unsigned char *' from smaller integer type 'int' [-Wint-to-pointer-cast]

Fix this by directly returning `NULL` in these functions.


noloader:
@MarcelRaad - I plan on taking this once I get it into a round of testing. My apologies for the late feedback.

noloader:
Thank you very much.

MarcelRaad:
Thanks for merging!

PR #175, opened 2016-06-01
cmake: add BUILD_(STATIC/SHARED) to specify which flavor to build

aberaud:
Make building the static of shared library optional.
Default is ON for both to preserve backward compatibility.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,8 @@ include(CheckCXXSymbolExists)
 # Settable options
 #============================================================================
 
+option(BUILD_STATIC "Build static library" ON)
+option(BUILD_SHARED "Build shared library" ON)
 option(BUILD_TESTING "Build library tests" ON)
 option(BUILD_DOCUMENTATION "Use Doxygen to create the HTML based API documentation" OFF)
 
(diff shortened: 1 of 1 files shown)


noloader:
Thank you very much.

noloader:
@aberaud - would it be possible to collect Cmake artifacts and put them in a `cmake` directory? Right now, it kind of pollutes the root directory:

	CMakeCache.txt
	CMakeFiles/
	CTestTestfile.cmake
	Makefile
	cmake_install.cmake
	cryptopp-config-version.cmake

noloader:
@aberaud,

Also, Cmake appears to hide its output, even with `V=1`:

$ make -f Makefile V=1
[  0%] Building CXX object CMakeFiles/cryptest-object.dir/bench1.cpp.o
[  1%] Building CXX object CMakeFiles/cryptest-object.dir/bench2.cpp.o
...

That's going to cause problems when we get support questions. Can you make it respond to `V=1` so it shows the command?

noloader:
@aberaud,

One more issue that needs attention. After building, the output artifacts are hidden from the user:

```
cryptopp$ gdb ./cryptest.exe
GNU gdb (Ubuntu 7.10-1ubuntu2) 7.10
./cryptest.exe: No such file or directory.
(gdb) 
```

We kind of need Cmake to stop hiding stuff from users.

aberaud:
Hi,
Thanks for merging fast :-)
CMake offers the possibility to build out of tree, just call it from the build directory :
```bash
cd cryptopp
mkdir build && cd build
cmake ..
# use CMakeLists.txt from the parent directory
make
# all build artifacts will be generated in the build directory, project directory will be left unchanged
```
Using a `build` subdirectory is common, but this can be done from any directory.

To build with verbose output using CMake, just add VERBOSE=1 after make :
```bash
make VERBOSE=1
```

Regards,
Adrien

noloader:
> CMake offers the possibility to build out of tree, just call it from the build directory...

My bad... I did not explain things well. We don't want users to do anything special. We want it to "just work" for them. When users have questions and have to do things, then it generates support questions.

We want to avoid support questions at nearly all costs. The best way I've found to do that is to ensure things "just works", even if it means we have to do extra stuff in the sources.

Issue #185, opened 2016-06-12
CBC_CTS_Encryption::ProcessLastBlock does not work inplace


berendeanicolae:
As CBC_CTS_Encryption inherits CBC_Encryption and CBC_Encryption::ProcessData allows the two buffers to be the same, I think CBC_CTS_Encryption::ProcessLastBlock should allow it too.
Or if not, an assert to check that are not the same would be useful.

noloader:
@berendeanicolae - Is this a bug or a feature? Do you have simple test program to duplicate the issue?

Are you experiencing a core dump with `make CXXFLAGS=-DDEBUG -g2 -O1`? Or are you seeing incorrect results? If its one of those two, then its definitely a bug.

Do you have a patch or diff to recommend?

berendeanicolae:
I would say it's a bug because it's not working as I expected. The documentation doesn't state that ProcessLastBlock should work inplace (it doesn't state anything regarding this issue). But regarding [ProcessData](https://www.cryptopp.com/docs/ref/class_stream_transformation.html#a26feabde21bc4d1783195969733e3bb0) it says "Either inString == outString, or they must not overlap".

Here is a code snippet which duplicates the issue http://pastebin.com/nFdN0U8B

noloader:
From the looks of it under GDB, it appears none of `tmp` is being encrypted (if I am parsing things correctly):

```
(gdb) x/70b src
0x7ad100:	0xbf	0x6c	0xd1	0xec	0x30	0x09	0xc3	0x46
0x7ad108:	0x6f	0xf1	0x88	0x8a	0x2a	0x31	0x0a	0xa1
0x7ad110:	0x10	0x11	0x12	0x13	0x14	0x15	0x16	0x17
0x7ad118:	0x18	0x19	0x1a	0x1b	0x1c	0x1d	0x1e	0x1f
0x7ad120:	0x20	0x21	0x22	0x23	0x24	0x25	0x26	0x27
0x7ad128:	0x28	0x29	0x2a	0x2b	0x2c	0x2d	0x2e	0x2f
0x7ad130:	0x30	0x31	0x32	0x33	0x34	0x35	0x36	0x37
0x7ad138:	0x38	0x39	0x3a	0x3b	0x3c	0x3d	0x3e	0x3f
0x7ad140:	0x5f	0xe6	0x94	0xd2	0xe0	0x22
(gdb) x/70b tmp
[...]

PR #186, opened 2016-06-12
fix bug #185

berendeanicolae:
Fix inplace ProcessLastBlock.

### DIFF
--- a/modes.cpp
+++ b/modes.cpp
@@ -207,9 +207,18 @@ void CBC_CTS_Encryption::ProcessLastBlock(byte *outString, const byte *inString,
 		// steal from next to last block
 		xorbuf(m_register, inString, BlockSize());
 		m_cipher->ProcessBlock(m_register);
-		inString += BlockSize();
-		length -= BlockSize();
-		memcpy(outString+BlockSize(), m_register, length);
+		if (inString == outString)
+		{
+			length -= BlockSize();
+			memcpy(outString, inString+BlockSize(), length);
+			memcpy(outString+BlockSize(), m_register, length);
+		}
+		else
+		{
+			inString += BlockSize();
+			length -= BlockSize();
+			memcpy(outString+BlockSize(), m_register, length);
+		}
 	}
 
 	// output last full ciphertext block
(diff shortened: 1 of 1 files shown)


modes.cpp
```diff
@@ -207,9 +207,18 @@ void CBC_CTS_Encryption::ProcessLastBlock(byte *outString, const byte *inString,
...
 		m_cipher->ProcessBlock(m_register);
-		inString += BlockSize();
-		length -= BlockSize();
-		memcpy(outString+BlockSize(), m_register, length);
+		if (inString == outString)
+		{
+			length -= BlockSize();
+			memcpy(outString, inString+BlockSize(), length);
+			memcpy(outString+BlockSize(), m_register, length);
+		}
```
noloader: When `inString == outString`, you have to use `memmove`, not `memcpy`.

Also see [Which one to use - memmove() or memcpy() - when buffers don't overlap?](http://stackoverflow.com/q/1960991/608639), [memcpy of overlapping buffers](http://stackoverflow.com/q/25629736/608639), [Glibc change exposing bugs](http://lwn.net/Articles/414467/) and [Strange sound on mp3 flash website](http://bugzilla.redhat.com/show_bug.cgi?id=638477) and

modes.cpp
```diff
@@ -267,11 +276,23 @@ void CBC_CTS_Decryption::ProcessLastBlock(byte *outString, const byte *inString,
...
-		xorbuf(outString, m_temp, m_register, BlockSize());
+		if (inString == outString)
+		{
+			memcpy(outString, inString+BlockSize(), length);
+			memcpy(outString+BlockSize(), m_temp, length);
+			// decrypt next to last plaintext block
+			memcpy(m_temp, pn1, length);
+			m_cipher->ProcessBlock(m_temp);
+			xorbuf(outString, m_temp, m_register, BlockSize());
+		}
```
noloader: Ditto.

berendeanicolae:
Fixed. Thanks - I've never heard of memmove before

noloader:
@berendeanicolae - I hope to start testing this soon. Its about number four on my list of things to do.

berendeanicolae:
@noloader - I think that the operations should performed in a SecByteBlock (a secure block of memory) and not directly in the buffer. I'll make some time in time weekend to fix this and run the tests suite on a linux machine.

berendeanicolae:
@noloader - I've modified ProcessLastBlock within CBC_CTS so that the stolenIV implementation is separated from the rest and now all the processing (encryption, xor, etc.) is done in a SecByteBlock.

I've run the tests from the pr ([pastebin](http://pastebin.com/3ubfLwGS)) on win x64 and they worked. Then I've run cryptest.sh on arch linux x64 and all the tests worked.

noloader:
@berendeanicolae,

I'm not sure what to do with this at the moment. I have not had the time to investigate it further.

Before anything happens, I'd like to get some test cases added. Test cases are a great first step. I think they are mandatory in this case to help us understand what's going on and to test the fix is a complete remdiation on all platforms.

I'm going to close this at the moment. I will probably ping you in the future to coordinate when I have the time to deep dive the issue.

Also, there could be a lurking underlying high-level policy/procedure issue. That issue is the in-place encryption and decryption. We may not have policy and procedures in place to deal with it. I hope that's not the case because the efficiencies gained with in-place transformations should not be taken for granted.

berendeanicolae:
@noloader 

So, would you like CBC_CTS to be patched to support in-place transformation?

It's true that in the documentation it's not specified that CBC_CTS supports in-place transformation but, as CBC_CTS is derived from CBC which supports in-place transformation CBC_CTS should also support it.

You said that you'd like some test cases to be added. I saw the tests are implemented in validat1.cpp. Do you want me to create some tests, or rather put the issue on hold for now?

PR #201, opened 2016-06-21
CMake cryptest binary is now named "cryptest.exe" on all platforms.

GamePad64:
As a result of our discussion, I am committing a piece of code, that makes "cryptest" binary named as "cryptest.exe" on non-Windows platforms to make it uniform with documentation.

Also, it adds a new target "cryptest.exe", so the library could be built using the [command from the documentation](https://www.cryptopp.com/wiki/Linux#Make_and_Install): `make static dynamic cryptest.exe`

This patch is tested on Ubuntu 12.04 and 16.04.


noloader:
Thank you very much.

PR #215, opened 2016-07-04
Add HMQV and merge untracked FHMQV

mouse07410:


mouse07410:
Added as a new branch "hmqv", so this PR is unnecessary

Issue #21, opened 2015-07-30
ECIES uses byte count, and not bit count, for label size


noloader:
This issue was originally investigated and reported by Jesse Wilson and Daniele Perito at [Problem with the way gfpcrypt HMAC's the encoding parameters' length in DHAES_MODE](https://groups.google.com/forum/#!msg/cryptopp-users/vR8GSL8wxPA/Bf9koUDyZ88J). Since that time, David Hook of the Bouncy Castle project investigated further and found another interop issue: Crypto++'s ECIES should use a bit count for the label, and not a byte count.

The issue was confirmed by cross referencing both P1363, revision 13 and subsequent P1363 2000 publications.

This report creates an actionable item for tracking.

DevJPM:
Does this actually change the interface (the user needs to supply bit
counts rather than byte counts?) or is it only some in-library only
thing needed for proper following of the specification?

BR

JPM

Am 30.07.2015 um 08:50 schrieb Jeffrey Walton:

noloader:
> Does this actually change the interface (the user needs to supply bit counts rather than byte counts?) 

Theoretically speaking, I don't believe it changes the interface. All we need to do is:

    diff --git a/gfpcrypt.h b/gfpcrypt.h
    index 0590254..1537b31 100644
    --- a/gfpcrypt.h
    +++ b/gfpcrypt.h
    @@ -454,7 +454,7 @@ public:
     		if (DHAES_MODE)
     		{
     			byte L[8] = {0,0,0,0};
    -			PutWord(false, BIG_ENDIAN_ORDER, L+4, word32(encodingParameters.size()));
    +			PutWord(false, BIG_ENDIAN_ORDER, L, word64(8 * encodingParameters.size()));
     			mac.Update(L, 8);
     		}
     		mac.Final(ciphertext + plaintextLength);
    @@ -483,7 +483,7 @@ public:
     		if (DHAES_MODE)
     		{
     			byte L[8] = {0,0,0,0};
    -			PutWord(false, BIG_ENDIAN_ORDER, L+4, word32(encodingParameters.size()));
    +			PutWord(false, BIG_ENDIAN_ORDER, L, word64(8 * encodingParameters.size()));
     			mac.Update(L, 8);
     		}
     		if (!mac.Verify(ciphertext + plaintextLength))

But in practice, it creates a lot of troubles because we should support:

  * Original behavior
  * Conforming behavior
  * Bouncy Castle interop for downlevel clients

[...]

PR #216, opened 2016-07-04
Make ECIES default compatible with BouncyCastle

mouse07410:


noloader:
Also see [Issue 21: ECIES uses byte count, and not bit count, for label size](http://github.com/weidai11/cryptopp/issues/21).

noloader:
@mouse07410,

What do you think about spinning up a ***ecies*** dev-branch, and then aligning `DLIES` and `ECIES` with Bouncy Castle?

mouse07410:
> What do you think about spinning up a ecies dev-branch, and then aligning DLIES and ECIES with Bouncy Castle?

No idea about DLIES, but for ECIES the patch is 3 lines (OK, maybe 6 - I don't remember), and it's been already tested with BC up to 1.55 and the current GitHub master. I don't want to start a whole separate branch bureaucracy for such a simple change (especially since it doesn't look like we've got a bunch of users banging on our doors demanding to preserve the old incompatible ECIES behavior). All my tests here are done with it#216 applied (because I don't have use for a library that's not compatible with BC :).

I think we should just merge #216 (the ECIES fix) into the master for 5.7, and be done with it.

noloader:
> I don't want to start a whole separate branch bureaucracy for such a simple change

Lol... That put milk up my nose.

Don't take those simple changes for granted :) I can't recount how many times I risked being tarred and feathered by users because a  simple change had unforeseen side effects :)

Also, I think we want to completely align with Bouncy Castle and Botan here. That is, we want things to "just work" for users. Performing the fix on ECIES means DLIES could have issues, so its not a complete remediation in security engineering parlance. We will definitly want to give DLIES its due attention.

I'll spin-up the ECIES branch to help with mods and testing. When you see the branch appear, please make a PR against it.

noloader:
Also see [Commit c3e2e0fb25fd](http://github.com/weidai11/cryptopp/commit/c3e2e0fb25fd1fa25e3a6e8eadb9ef78e8158be9).

PR #241, opened 2016-08-07
Fixed install path for Windows in CMake

GamePad64:
MinGW produces two files for a shared target: import library and shared target. Import library must be located in /lib (or its Windows analogue), and shared library must be located in /bin.


noloader:
Thank you very much.

Issue #83, opened 2015-12-14
zdeflate error 


srwinohio:
When I compress a plain
[atom.txt](https://github.com/weidai11/cryptopp/files/61198/atom.txt)

 text buffer, I get an assert from zdeflate in MatchFound, line 
assert(length >= 3 && length < COUNTOF(lengthCodes));

Code snippet that I am using to compress......

	CryptoPP::Gzip zipper;

         //is -  is an ifstream passed in

	//read the source file
	if (!is.is_open() || !os.is_open() )
	{
		return(-1);
	}

	// get its size:
	long fileSize = getFileSize(is);

	//allocate a buffer big enough to read the entire file
	byte *buffer = new byte[(long)fileSize+1024];

	if (buffer != NULL)
	{
		//read the entire file into the buffer
		is.read((char *)buffer, fileSize);

		//compress it
		zipper.Put((byte*)buffer, (int)fileSize);
		zipper.MessageEnd();
      }

noloader:
@srwinohio - ACK. I'm going to need some time to look at it.

Can you add the code for `getFileSize`?

srwinohio:
unsigned long CCactCompression::getFileSize(ifstream *ifs)
{
	// get its size:
	streampos fileSize;

	ifs->seekg(0, ios::end);
	fileSize = ifs->tellg();
	ifs->seekg(0, ios::beg);

	return((unsigned long)fileSize);
}

This appears to have something to do with the source data. The file I test with is a repeated string. If I use a more "random" type file, for instance a log file from an application, it works just file.

Steve

noloader:
> This appears to have something to do with the source data. 

Thanks Steve.

I'm actually somewhat familiar with this issue. I've seen it show up on occasion under Valgrind. I was chasing it for months. I added the asserts in an attempt to locate the issue.

I had a hard time duplicating it in the past. I *thought* we cleared the actual memory issue based on a Coverity finding/CID 148082 (https://scan.coverity.com/projects/cryptopp). Though its not readily apparent, the `SecBlock` is really an array of [`CodeInfo`](https://www.cryptopp.com/docs/ref/zinflate_8h_source.html#l00053) objects.

If you can, would you build the library with `g2` and `-O0` or `-O1`, and then run the test case under Valgrind? Valgrind requires [`-O0` or `-O1` for accurate results](http://valgrind.org/docs/manual/quick-start.html). You can do so with:

    make distclean
[...]

PR #242, opened 2016-08-10
Core dump in HuffmanEncoder::GenerateCodeLengths in case of only one leaf in tree.

kvirund:


noloader:
Thanks @kvirund. Do you have  small test case for the crash? I'd like to add it to out test suite.

kvirund:
Unfortunately I cannot share that data, but I will try to build another test case.

kvirund:
Actually, I can share values of arguments passed into `HuffmanEncoder::GenerateCodeLengths(unsigned int *codeBits, unsigned int maxCodeBits, const unsigned int *codeCounts, size_t nCodes)` when core dump occurred:

```
nCodes = 0x1e
codeCounts = {1, 0, 0 x 27 times, 0}
maxCodeBits = 15
codeBits = <uninitialized>
```

kvirund:
Essentially it is something like:

```
#include <gzip.h>

  #include <iostream>

  int main()
  {
      const size_t nCodes = 0x1e;
      const unsigned int codeCounts[nCodes] = {1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
      const unsigned int maxCodeBits = 15;
      unsigned int codeBits[nCodes] = { -1 };

      CryptoPP::HuffmanEncoder::GenerateCodeLengths(codeBits, maxCodeBits, codeCounts, nCodes);
      for (int i = 0; i < nCodes; ++i)
      {
          std::cout << codeBits[i] << std::endl;
      }

      return 0;
  }
```

On Linux it works correctly but on Windows in Debug mode at line `const size_t depth = STDMIN(maxCodeBits, tree[n].depth + 1);` `n` had wrong value. Because it has value `tree[i].parent` where `i` is `treeBegin` where `treeBegin` points to the single symbol. But because this symbol is single it does not have parent (it was not merged with any other symbols in loop `for (i=nCodes; i<tree.size(); i++)`, because `tree.size()` is equal to `nCodes`: `tree.resize(nCodes + nCodes - treeBegin - 1);` because `1 == nCodes - treeBegin` => `nCodes == nCodes + nCodes - treeBegin - 1`).

noloader:
@kvirund,

> ... now I cannot reproduce it from scratch even on Windows...

Does that mean something else changed (like you started working from Master) and the issue is no longer present? Or does that mean the issue still exists, but the test case does not capture it? Or maybe something else?

I guess the second question is, do we still need to consider the pull request?

kvirund:
No, issue still exists. Just when I tried to reproduce it I built latest version of CryptoPP and in this version variables (`tree` specifically) were initialized by zeros. So I was not able to reproduce the issue just because values became lucky. But error is still there. Because code is the same as where I got core dump.

I got value `tree[0xcccccccc00000000]` because index had gotten from `const size_t n = tree[i].parent;`. But parent is the union: `union {size_t parent; unsigned depth, freq;};`. So when last element was initialized on line `zdeflate.cpp:143`: `tree[tree.size()-1].depth = 0;`, it initialized only low part of `parent` because size of `unsigned` is half shorter than size of `size_t`.

Code that I mentioned is the reproducible case. Just I don't know how to compile project so that tree is initialized by `0xcccccccccccccccc`. But to check my theory you can do it manually (just add before line `zdeflate.cpp:119` the following code `tree[i].parent = 0xccccccccccccccccull;`. If algorithm would have worked correctly it should not mean anything).

noloader:
> I got value `tree[0xcccccccc00000000]` because index had gotten from const size_t n = tree[i].parent;. But parent is the union: `union {size_t parent; unsigned depth, freq;}` ...

Ah, OK, thank you very much. I did not realize a union was in play. Related, in C++, its [undefined behavior to read from the inactive union member](http://stackoverflow.com/q/11373203). In practice nearly every (every?) C++ compiler does the right thing because its legal to do in C.

I've seen that before under MSVC++. I encountered it when trying to avoid an unitialized warning for `LARGE_INTEGER` in `hrtimer.cpp`. In that case, can you test this:

```
$ git diff
diff --git a/zdeflate.cpp b/zdeflate.cpp
index a016b30..54135ad 100644
--- a/zdeflate.cpp
+++ b/zdeflate.cpp
@@ -91,7 +91,7 @@ struct HuffmanNode
 {
        // Coverity finding on uninitalized 'symbol' member
        HuffmanNode()
-               : symbol(0), parent(0) {}
+               : symbol(0) { depth=0, freq=0; }
        HuffmanNode(const HuffmanNode& rhs)
                : symbol(rhs.symbol), parent(rhs.parent) {}
```

kvirund:
Actually the problem is not in the initializing. Trash in high part of `parent` value just uncovers the problem.

If I understand right, algorithm should build `tree` that reflects the Huffman codes. So, in the loop at `zdeflate.cpp:132` it builds that tree by creating parent node over two nodes with least frequency values. The case I stepped on is when `tree` initially has only one node. So **no any parent nodes will be built**. And this single node won't have a parent.

After that, algorithm recursively counts depth of each node at `zdeflate.cpp:150`. Here "recursively" means that to count depth of some node, algorithm takes `depth` value of its parent node. So body of this loop **expects that each node being processed in this loop has some parent node**. But it is not true when only the single node presented in the tree. That is the problem.

noloader:
@kvirund,

I added the test case at [Commit 1e7c837442231a55](http://github.com/weidai11/cryptopp/commit/1e7c837442231a55fab2daa39c353755e1023bc6). Thank you very much.

The test case does not trigger on OS X or Linux, so I don't know if the reworked `HuffmanEncoder` ctors actually solve the problem on Windows. (And there could be two different problems here).

kvirund:
Do you need something else from me?

noloader:
> Do you need something else from me?

Well, I'm not sure.

I tested [Commit 1e7c837442231a55](http://github.com/weidai11/cryptopp/commit/1e7c837442231a55fab2daa39c353755e1023bc6) on Windows 7 with Visual Studio 2010 under both X86 and X64. I can't seem to reproduce the issue.

What platform and compiler were you experiencing the issue on?

kvirund:
It was Microsoft Visual Studio 2015, Windows 7. But I did not use CMake scripts that comes with library.

Actually, I think core dump will not be reproduced anymore. But don't you agree with my resonings? I mean my notice that the loop at `zdeflate.cpp:150` work in assumption that each node in range [treeBegin, nCodes) has a parent node. But for my test case we have only one node that does not have parent node. So that is the mistake in the logic: it does not handle properly corner case.

noloader:
> Actually, I think core dump will not be reproduced anymore.

In that case, I would prefer not take action. I want to avoid inadvertently breaking something with an unintended consequence.

After AES and SHA, the compressors and decompressors are some of the hairiest code in the library. In [Issue 83: zdeflate error](http://github.com/weidai11/cryptopp/issues/83) I knocked something loose while chasing an uninitialized bug that surfaced under Valgrind. It took some time for the break to surface and subsequent fix, and I don't want to repeat it.

I'm not married to the "no action" position. If I can get hold of a test case that reproduces the issue, then I'm happy to move against it.

PR #258, opened 2016-09-07
Fix for CAST256 - Key-size

Blacktempel:
Returned valid size is 16+1+1+1+1+1.... instead of 16+4... (16, 20, 24, 28, 32).
This fixes it.


noloader:
Thank you very much.

PR #260, opened 2016-09-08
shortened ChaCha typedefs

DevJPM:
shortened ChaCha typedefs because they are essentially the same twice and this is more library-standard


noloader:
Thank you vry much.

PR #263, opened 2016-09-11
Changes for BouncyCastle interoperability

mouse07410:
Also removed the "deprecation" warning, because this change is what that warning was telling about.


noloader:
Nice, thanks.

PR #270, opened 2016-09-13
Fix clang warnings about undefined variable templates in pkcspad.h

DimitryAndric:
Recent versions of clang warn about missing `PKCS_DigestDecoration` definitions, specifically the decoration members.  For example:

    In file included from testpkcs.cpp:8:
    ./pkcspad.h:89:53: warning: instantiation of variable 'CryptoPP::PKCS_DigestDecoration<CryptoPP::SHA1>::decoration' required here, but no definition is available [-Wundefined-var-template]
                                    return HashIdentifier(PKCS_DigestDecoration<H>::decoration, (size_t)PKCS_DigestDecoration<H>::length);
                                                                                    ^
    ./pubkey.h:608:19: note: in instantiation of member function 'CryptoPP::PKCS1v15_SignatureMessageEncodingMethod::HashIdentifierLookup::HashIdentifierLookup2<CryptoPP::SHA1>::Lookup' requested here
            return L::Lookup();
                      ^
    ./pubkey.h:638:26: note: in instantiation of member function 'CryptoPP::TF_ObjectImplBase<CryptoPP::TF_VerifierBase, CryptoPP::TF_SignatureSchemeOptions<CryptoPP::TF_SS<CryptoPP::PKCS1v15, CryptoPP::SHA1, CryptoPP::RSA, int>,
          CryptoPP::RSA, CryptoPP::PKCS1v15_SignatureMessageEncodingMethod, CryptoPP::SHA1>, CryptoPP::RSAFunction>::GetHashIdentifier' requested here
    class CRYPTOPP_NO_VTABLE TF_ObjectImpl : public TF_ObjectImplBase<BASE, SCHEME_OPTIONS, KEY_CLASS>
                             ^
    ./pkcspad.h:32:20: note: forward declaration of template entity is here
            static const byte decoration[];
                              ^

This is because pkcspad.h doen't contain explicit instantiation declarations for `PKCS_DigestDecoration`.  These warnings can be fixed by adding them in the `!CRYPTOPP_IS_DLL` case, which appears to apply only to Windows.

### DIFF
--- a/pkcspad.h
+++ b/pkcspad.h
@@ -54,6 +54,16 @@ CRYPTOPP_DLL_TEMPLATE_CLASS PKCS_DigestDecoration<SHA224>;
 CRYPTOPP_DLL_TEMPLATE_CLASS PKCS_DigestDecoration<SHA256>;
 CRYPTOPP_DLL_TEMPLATE_CLASS PKCS_DigestDecoration<SHA384>;
 CRYPTOPP_DLL_TEMPLATE_CLASS PKCS_DigestDecoration<SHA512>;
+#else
+extern template class PKCS_DigestDecoration<SHA1>;
+extern template class PKCS_DigestDecoration<RIPEMD160>;
+extern template class PKCS_DigestDecoration<Tiger>;
+extern template class PKCS_DigestDecoration<SHA224>;
+extern template class PKCS_DigestDecoration<SHA256>;
+extern template class PKCS_DigestDecoration<SHA384>;
+extern template class PKCS_DigestDecoration<SHA512>;
+extern template class PKCS_DigestDecoration<Weak1::MD2>;
+extern template class PKCS_DigestDecoration<Weak1::MD5>;
 #endif
 
 //! <a href="http://www.weidai.com/scan-mirror/sig.html#sem_PKCS1-1.5">EMSA-PKCS1-v1_5</a>


noloader:
Thanks @DimitryAndric,

I'm working another problem at the moment, so I don't have time to look at it.

Off the top of my head, I thought `dll.cpp` provides the definitions for some of them. *"some of them"* means the ones used in the [FIPS DLL](http://www.cryptopp.com/wiki/FIPS_DLL).

> Recent versions of clang ...

What versions of Clang are warning? We should be detecting capturing it in our test script, but I don't recall seeing it.

DimitryAndric:
Some definitions are in dll.cpp (SHA1 through SHA512), and some are in pkcspad.cpp (MD2, MD5, RIPEMD160 and Tiger).  There is probably a reason for this split, but I don't know it. :)

That said, the definitions in the cpp files are something different from a template instantiation declaration (see http://en.cppreference.com/w/cpp/language/class_template) in the headers.  These ensure that the compiler isn't going to instantiate any implicit version of the template.

Clang 3.9.0 and higher are now emitting warnings for this.  I encountered it during a FreeBSD ports build with clang 3.9.0, where a program called clementine-player (an audio player) was using cryptopp headers, and got errors similar to what I described above: http://package18.nyi.freebsd.org/data/headi386PR212343-default/2016-09-11_07h04m19s/logs/errors/clementine-player-1.3.1.log

For now we worked around it by silencing the warnings explicitly in the clementine-player build, but it would be nicer to fix it in the cryptopp headers.

noloader:
@DimitryAndric,

> Some definitions are in dll.cpp (SHA1 through SHA512), and some are in pkcspad.cpp (MD2, MD5, RIPEMD160 and Tiger). There is probably a reason for this split, but I don't know it. :)

Yes, its the awful [Windows FIPS DLL](http://cryptopp.com/wiki/FIPS_DLL). FIPS gear goes in dll.cpp, while non-FIPS can go just about anywhere.

I was able to duplicate the warnings under MacPorts Clang 3.9 on OS X.

> it would be nicer to fix it in the cryptopp headers.

Yes, absolutely. We get this PR merged after I can get it into a testing cycle. Give me a day or two.

noloader:
Thank you very much. Now the real testing begins...

noloader:
My apologies... this broke OS X immediately so I had to revert it. See the comments at [PR 294](http://github.com/weidai11/cryptopp/pull/294).

DimitryAndric:
Right, a rather lame fix for this would be to surround the extern template declarations with `#ifdef __clang__`.  But it is probably better to figure out why gcc chokes on it.  I tried several versions of gcc on FreeBSD, but they all seemed to be able to compile it just fine.  I will have a look at what happens on OSX.  Which gcc version were you using on OSX?

noloader:
@DimitryAndric,

The `g++` was actually on of those linked-to-Clang-behind-the-scenes ones:

```
$ g++ --version
Configured with: --prefix=/Applications/Xcode-5_1.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 5.1 (clang-503.0.40) (based on LLVM 3.4svn)
Target: x86_64-apple-darwin12.6.0
Thread model: posix
```

And `c++`, which is what we usually use on OS X:

```
$ c++ --version
Apple LLVM version 5.1 (clang-503.0.40) (based on LLVM 3.4svn)
Target: x86_64-apple-darwin12.6.0
Thread model: posix
```

noloader:
@DimitryAndric, This is turning out to be a thorny problem. I created a dev-branch called ***clang*** to work it so we can commit and test without tainting Master.

noloader:
@DimitryAndric, This is turning out to be a thorny problem. I created a dev-branch called ***clang*** to work it so we can commit and test without tainting Master.

I'm beginning to wonder if this is a Clang diagnostic bug or really a problem. I can't get another compiler to produce it. And when I appease Clang with a definition, it results in multiple definitions during link. Finally, Clang 3.9 only complains about `PKCS_DigestDecoration<H>::decoration` (an array), and not `PKCS_DigestDecoration<H>::length` (a scalar).

We are also going to track this with a bug report. These changes have the potential to cut deep, so we need the audit trail.

noloader:
@DimitryAndric, The ***clang*** dev-branch is ready for testing. It was tricky to get right due to the interactions among the platforms. It took three commits to get right. Also see [Issue 300, Comment 249043613](https://github.com/weidai11/cryptopp/issues/300#issuecomment-249043613).

noloader:
@DimitryAndric, We just finished the Clang testing and it was OK. I also ran [our comprehensive test suite](https://github.com/weidai11/cryptopp/blob/master/cryptest.sh) on NetBSD 7.0. It tested 41 different configurations and the tests results were OK.

Since the testing was successful, I deleted the *clang-dev* branch. I probably should have left it open since Clang has so many problems...

Issue #257, opened 2016-09-07
SHA3 HMAC Support


kinap:
I know the Keccak [site ](http://keccak.noekeon.org/) says HMAC is not necessary for SHA3 MAC construction. However, we have an environment where SHA3 and HMAC may be combined so I hacked in support for an internally maintained fork, something I'm trying to get away from.

Before cleaning this up and submitting a PR I wanted to have a discussion on the desirability of this. I seem to remember some resistance to the idea, but can't find an Issue or a PR for it.

noloader:
@kinap,

Email me off-list and I'll loop you in on some private emails. My email address is *noloader, gmail account*.

Also note HMAC for SHA-3 is relatively new. The idea has been around as long as the hash itself, but its been formalized recently (so things can interop). Also see [RFC 6979, Deterministic Signatures for DSA/ECDSA using SHA3?](http://www.ietf.org/mail-archive/web/saag/current/msg06894.html) on the SAAG mailing list.

noloader:
@kinap,

We are getting ready to make some changes in this area. First, we added a `Keccak` class tht provides the former NIST behavior, before FIPS 202 was finalized. Second, we changed SHA3 to conform to FIPS 202. Both changes are sitting in the [Keccak dev-branch](http://github.com/weidai11/cryptopp/tree/keccak).

Related to (2), we also changed the Algorithm Name of SHA3 from the string `SHA-3-NNN` to `SHA3-NNN`. There was an extra hyphen that was less confirming.

We are still up in the air with respect to sponge functions (other than F1600) and XOFs. We don't really have an idea of how they should look at the design and implementation level for software. (I have some ideas in the back of my head). However, perfection should not be the enemy of the good, so we pulled the trigger and provided Keccak.

Nothing has changed with respect to your request for HMAC or KMAC. However, some hurdles have been cleared that should make it easier on folks.

noloader:
@kinap, @AndrewWPhillips,

There's a thread on the mailing list at [Keccak dev-branch](http://groups.google.com/forum/#!topic/cryptopp-users/hhXiNXTQFEI). If you have feedback, then please provide it. There are plans to merge the Keccak branch shortly.

The imperativeness is due to the impending Crypto++ 5.6.4 release.

noloader:
Also see [Issue 262: Need better design for Keccak](https://github.com/weidai11/cryptopp/issues/262).

DevJPM:
Note sure if this is clear to everyone:

[...]

PR #280, opened 2016-09-16
fixed Keccak and SHA3 to support HMAC

DevJPM:
added the blocksize constant and member functions to Keccak and SHA3
(and all derivatives) as well as some compile-time-checks

### DIFF
--- a/keccak.h
+++ b/keccak.h
@@ -56,6 +56,8 @@ public:
 	void Restart();
 	void TruncatedFinal(byte *hash, size_t size);
 
+	unsigned int BlockSize() const { return r(); }
+
 protected:
 	inline unsigned int r() const {return 200 - 2 * m_digestSize;}
 
@@ -70,10 +72,14 @@ class Keccak_224 : public Keccak
 {
 public:
 	CRYPTOPP_CONSTANT(DIGESTSIZE = 28)
+	CRYPTOPP_CONSTANT(BLOCKSIZE = 200 - 2 * DIGESTSIZE)
 
 	//! \brief Construct a Keccak-224 message digest
 	Keccak_224() : Keccak(DIGESTSIZE) {}
 	CRYPTOPP_CONSTEXPR static const char *StaticAlgorithmName() {return "Keccak-224";}
+private:
+	CRYPTOPP_COMPILE_ASSERT(BLOCKSIZE < 200); // ensure there was no underflow in the math
+	CRYPTOPP_COMPILE_ASSERT(BLOCKSIZE > DIGESTSIZE); // this is a general expectation by HMAC
 };
 
 //! \class Keccak_256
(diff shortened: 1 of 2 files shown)


DevJPM:
This is the fix for the SHA-3 / Keccak with HMAC issue I promised to deliver after 5.6.4 in Issue #257

noloader:
Sorry about not commenting sooner. This fell off the radar. Do you have plans to make `r()` dynamic or runtime configurable? If not, do you see any problems with tweaking:

```
unsigned int BlockSize() const { return r(); }
```

To:

```
unsigned int BlockSize() const { return BLOCKSIZE; }
```

noloader:
The second comment is, iterated hashes based on DM-constructions have block sizes. Should we [publicly?] overload `BlockSize()` for a sponge construction:

```
unsigned int BlockSize() const { ... }
```

To toss an alternative out there:

```

public:
    unsigned int SpongeSize() const { ... }

protected:
    unsigned int BlockSize() const { ... }
```

For the classes that need access to operate the cipher, we can use friendship.

Or is this just overkill?

DevJPM:
> Do you have plans to make `r()` dynamic or runtime configurable?

If we plan to allow for dynamic permutation-selection, then yes, we have to make this one dynamic. However I'd much prefer making the permutation a template parameter (at least for the instantiation) if possible. And the second reason to delegate to r() was to push the instantiation into the base class, so we don't have to place the BlockSize() definition in _all_ derived classes (because BLOCKSIZE is unknown at the upper level).

As for hiding the block-size, I don't really think this is a good idea as it's the library standard not to hide it and it would kinda break the "contract" that blocksize is guaranteed to be accessible by anybody. If you feel better, we could indeed add a SpongeSize() / BitRate() / AbsorptionRate() function to clarify the meaning of BlockSize (inline-defaulting to BlockSize())

noloader:
> If we plan to allow for dynamic permutation-selection, then yes, we have to make this one dynamic. However I'd much prefer making the permutation a template parameter

OK, good. It also disconnect `r()` from `CRYPTOPP_CONSTANT(BLOCKSIZE = 200 - 2 * DIGESTSIZE)` :)

Do you feel like making the change? I'll get it merged shortly.

I'm also going to merge @MarcelRaad's changes to the VS project files after getting an issue documented for it (change control; not a bug). The issue report will allow us to back it out quickly by making it easy to locate. Take a look over the diff's he provided at https://github.com/weidai11/cryptopp/pull/286 and https://patch-diff.githubusercontent.com/raw/weidai11/cryptopp/pull/286.diff .

DevJPM:
> Do you feel like making the change?

I'm kinda confused right now which change you mean? Moving the BlockSize functions into the derived classes and redefining them to be BLOCKSIZE?

noloader:
> Moving the BlockSize functions into the derived classes and redefining them to be BLOCKSIZE?

My bad... This one:

```
CRYPTOPP_CONSTANT(BLOCKSIZE = 200 - 2 * DIGESTSIZE)
unsigned int BlockSize() const { return BLOCKSIZE; }
```

Or is that wrong?

DevJPM:
@noloader updated.

noloader:
Thank you very much.

noloader:
Also se [Commit de04771aa8af7ec8](http://github.com/weidai11/cryptopp/commit/de04771aa8af7ec8afdbd47af3a5773f553579bf).

PR #281, opened 2016-09-16
Restore MARS capability to use 448-bit keys

mouse07410:
There is no reason for MARS to be limited to 256-bit keys, and some applications (e.g., https://github.com/cryfs/cryfs.git) request larger keys.

This PR restores MARS 448-bit key capability.


noloader:
That's a tad bit lite on testing, isn't it :)

Do we have test vectors for 448-bit? I'm guessing not since the validation suite or test vectors probably should have crashed with the over-sized key.

PR #286, opened 2016-09-19
Trivial VC10 project fixes

MarcelRaad:
- removed tabs and newlines as Visual Studio always deletes them when saving the projects
- removed per-file optimization properties which made changing the property unnecessarily hard
- removed per-file PreprocessorDefinitions properties which had no effect

This makes working with the project files much easier if you have additional configurations or different properties.


noloader:
Thanks Marcel. Give me a few minutes to test these.

There's probably a lot of extra cruft if you look closely. Its been carried around for years. I do my best to remove things like VC7.1/VS2003 compatibility when the targeted tool chain is VS2010.

At one time I was experimenting with speeding up the the project loads by folding different configurations into a single rule. It was time consuming, and I think Visual Studio unrolled them after it read the project file.

MarcelRaad:
@noloader Hmm, only one file, cryptlib.vcxproj, should be affected by the two commits thst GitHub won't display. I'm using the TortoiseGit GUI under Windows, which displays the changes correctly with "show log".

The command "git log -p -3" seems to do the same thing after checking out the vc10_projects branch, but the diff format doesn't look as nice (huge blocks of additions and removals instead of inline diffs). Perhaps it's easiest to compare cryptlib.vcxproj of that branch with the one from the master branch using your favorite diff tool?

noloader:
@MarcelRaad,

It looks like you did that folding. I think its a good idea in general.

I think the way to approach this is to open a Bug Report. That gives us the audit trail we need. I'll also cite the commits to make it easy to back out in case something goes terribly wrong.

I'll handle the bug report and then merge.

Thanks for the contrib.

noloader:
Thank you very much.

MarcelRaad:
Thank you for merging so quickly!

noloader:
@MarcelRaad, We performed the same on `cryptest.vcxproj`. See [Commit b4f08a3fa17fe652](http://github.com/weidai11/cryptopp/commit/b4f08a3fa17fe652b2e72e33ca18e48bf9972f5c).

Issue #292, opened 2016-09-20
Commit ccd59710 broke Keccak


mouse07410:
Current master fails to compile. Keccak got broken by commit ccd59710 (was it tested before merging?):
```
. . . . .
clang++ -maes -mpclmul -mrdrnd -msse2 -mssse3 -msse4.2 -mtune=native -Os -Ofast -std=c++11 -fPIC -march=native -pipe -c keccak.cpp
In file included from keccak.cpp:18:
./keccak.h:81:40: error: constexpr function's return type 'std::string' (aka 'basic_string<char,
      char_traits<char>, allocator<char> >') is not a literal type
        CRYPTOPP_CONSTEXPR static std::string StaticAlgorithmName() { return "Keccak-" + IntToStri...
                                              ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string:4271:48: note: 
      'basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >' is not literal because
      it is not an aggregate and has no constexpr constructors other than copy or move constructors
_LIBCPP_EXTERN_TEMPLATE(class _LIBCPP_TYPE_VIS basic_string<char>)
                                               ^
In file included from keccak.cpp:18:
./keccak.h:82:34: error: virtual function cannot be constexpr
        CRYPTOPP_CONSTEXPR unsigned int BlockSize() const { return BLOCKSIZE; }
                                        ^
./cryptlib.h:984:23: note: overridden virtual function is here
        virtual unsigned int BlockSize() const {return 0;}
                             ^
2 errors generated.
make: *** [keccak.o] Error 1
```

```
$ git blame keccak.h
. . . . .
cd59710 (DevJPM         2016-09-20 11:27:25 +0200  80)         Keccak_Final() : Keccak(DIGESTSIZE) {}
ccd59710 (DevJPM         2016-09-20 11:27:25 +0200  81)         CRYPTOPP_CONSTEXPR static std::string StaticAlgorithmName() { return "Keccak-" + IntToString(DIGESTSIZE * 8); }
. . . . .
```

Mac OS X 10.11.6, Xcode-8.0.

noloader:
Reverted at [Commit c298dfdbb017076d](http://github.com/weidai11/cryptopp/commit/c298dfdbb017076de1e2f1b6bb8ccd00fe85fb95).

[...]

Issue #288, opened 2016-09-19
Making the derived SHA3 and Keccak classes templates with typedefs instead of their own classes?


DevJPM:
While working on Skein I used the idea that the final, main instances only differentiate in digest-size.

So I wrote a template class, with the digest-size as a template and generalized everything digest-size dependent (e.g. the constants, the algorithmName(),...).

Should we also consider this for Keccak / SHA3?

The benefit would be that adding a new compile-time digest Keccak instance would come down to `typedef Keccak<16> Keccak_128;` instead of the 10-20 lines construct we have now, which would greatly increase maintainability and would probably save us about 100 lines or so of code.

Note: I consider this to be a really low priority issue.

DevJPM:
What I proposed is now PR #291

PR #296, opened 2016-09-21
Templated Keccak and SHA3

DevJPM:
templated Keccak and SHA3 to reduce code-size, added a
StaticAlgorithmName() to the base classes and restricted use of
constexpr to this new function in the base classes

Addresses issues #288 and #292 as proposed in a private mail from @noloader 

It built successfully under VS2015, no other compilers tested (as of now).


noloader:
Tested OK under OS X. Thank you very much.

PR #301, opened 2016-09-23
When calculating the AES block cipher, allocate 4K of memory on the stack

johnwbyrd:
When calculating the AES block cipher, allocate 4K of memory on the stack instead of 256+ bytes.  Search within that 4K space to put the 256-byte aligned Locals struct in a place which does not have 4K cache conflicts with the `Te` temporary buffer.   This permits us to call `_malloca()` or `alloca()` once per call of this function.

This commit also makes sure that the Microsoft-only `_freea()` occurs at the correct location instead of at a pointer to the middle of the stack, when the memory allocated by `_malloca()` or `alloca()` is not 256-byte aligned.

### DIFF
--- a/rijndael.cpp
+++ b/rijndael.cpp
@@ -1035,6 +1035,9 @@ void Rijndael_Enc_AdvancedProcessBlocks(void *locals, const word32 *k);
 
 #if CRYPTOPP_BOOL_X64 || CRYPTOPP_BOOL_X32 || CRYPTOPP_BOOL_X86
 
+/* Determine whether the range between begin and end overlaps
+ * with the same 4k block offsets as the Te table.
+ */
 static inline bool AliasedWithTable(const byte *begin, const byte *end)
 {
 	size_t s0 = size_t(begin)%4096, s1 = size_t(end)%4096;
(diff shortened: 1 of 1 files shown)


noloader:
Thanks John. Give me a few hours for some offline testing.

noloader:
Thanks again @johnwbyrd. As a first step in remediating the [302 issue](https://github.com/weidai11/cryptopp/issues/302), I backed out the portion of the [823 commit](http://github.com/weidai11/cryptopp/commit/823bc93357da32a3a4a2b71b9915a4e124839d18) that introduced this bug. We are now back at where things were before the break.

I think we are going to pass on the [301 patch](http://github.com/weidai11/cryptopp/pull/301) at the moment (but thank you very much for it). I think your critique was fair, but I can't justify doing the same thing after we were criticized for it.

dirvine:
Yes, thank you very much @johnwbyrd for the report and solution. I agree with Jeff here in backing out the code that introduced this.

johnwbyrd:
I am sorry, but I have some bad news.  **Simply removing _freea from the code does not fix the underlying bug.**  This bug should be re-opened.

The proposed fix makes two important, incorrect assumptions.  From the documentation at https://msdn.microsoft.com/en-us/library/5471dc8s.aspx :

> _malloca allocates size bytes from the program stack or the heap if the request exceeds a certain size in bytes given by _ALLOCA_S_THRESHOLD. The difference between _malloca and _alloca is that _alloca always allocates on the stack, regardless of the size. Unlike _alloca, which does not require or permit a call to free to free the memory so allocated, _malloca requires the use of _freea to free memory. In debug mode, _malloca always allocates memory from the heap.

I confirm that the implemented Microsoft code actually corresponds to this behavior.  This behavior implies several problems in the fix.

1. If the code is compiled in debug mode, _alloca always allocates memory on the heap, not the stack.  That means that in debug mode, calling _malloca without a corresponding _freea is guaranteed to leak memory.  And it's AES cipher block data, which is exactly the kind of memory you shouldn't be leaking.

2. If the memory requested is higher than _ALLOCA_S_THRESHOLD, then the allocation always occurs on the heap.  As of this writing_ALLOCA_S_THRESHOLD happens to be 1024.  The memory currently allocated is a maximum of 255 + sizeof( Locals ), which currently happens to be around 583 or so.  So this structure is allocated on the stack in release mode.  However, if either the _ALLOCA_S_THRESHOLD setting changes, or the size of the Locals variable grows to be 768 bytes or larger, then _malloca() will allocate on the heap, causing the memory leak as described above.  This specific item could be checked with an assert, but even if it is, it won't fix item 1 above.

Ergo, I claim that the minimal change strategy will not fix this bug.  Again, I recommend going with my original patch, which calls _malloca and _freea in correct pairs in all cases.  

The performance impact of the recommended patch is that exactly one heap allocation and free will occur during the decoding of one block.  This is not optimal, but you really need to get your code correct before worrying about performance.

noloader:
Thanks @johnwbyrd.

We've been discussing this on a sidebar since we reverted. I was concerned there was underlying lingering issue like this. One of the primaries did not feel comfortable either, and he advised taking the patch. Given your additional finding's and DB's position, how about resubmitting the patch.

And its awful the issue went undetected for so long. Rijndael/AES will be a top priority next go around. We are going to get some logical separation of components so we can maintain that file.

Thanks again.

johnwbyrd:
I've resubmitted the patch.  Please have some fresh eyes look at it carefully and consider the issues above, or else we'll be patching this patch forever.

PR #306, opened 2016-09-24
Incorrect calls to _malloca() to _freea() can leak memory and/or cause crashes on Windows

johnwbyrd:
Allocate 4K of memory using _malloca instead of 256+ bytes.  Search within that 4K space to put the 256-byte aligned Locals struct in a place which does not have 4K cache conflicts with the Te temporary buffer.   This permits us to call _malloca() or alloca() once per call of this function.  This commit also makes sure that the Microsoft-only _freea() occurs at the correct location instead of at a pointer to the middle of the stack, when the memory allocated by _malloca() or alloca() is not 256-byte aligned.


noloader:
Thanks John. merged.

PR #308, opened 2016-09-26
Fix Cmake (Issue 273)

FloriansGit:
Make use of `CHECK_CXX_COMPILER_FLAG()` for `-march=` availability checks.

And check that no architecture was predefined by the environment or the user.


noloader:
Thanks you very much.

PR #309, opened 2016-09-27
Fixes #276

### LINKED ISSUE #276: CMake produces artifacts without debugging information
Sigh...

```
cryptopp-build$ rm -rf * && CXX=/opt/solstudio12.2/bin/CC cmake ../cryptopp
cryptopp-build$ make
Scanning dependencies of target cryptopp-object
[  1%] Building CXX object CMakeFiles/cryptopp-object.dir/cryptlib.cpp.o
CC: Warning: -xchip=native detection failed, falling back to -xchip=generic
[  2%] Building CXX object CMakeFiles/cryptopp-object.dir/cpu.cpp.o
CC: Warning: -xchip=native detection failed, falling back to -xchip=generic
...
```

And then

```
(dbx) where
=>[1] CryptoPP::Baseline_Multiply2(0x23e64d0, 0x240b6f0, 0x240b7b0, 0x240b7b0, 0xa, 0x23dcbc8), at 0x182fac7 
  [2] CryptoPP::RecursiveMultiply(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x1846067 
  [3] CryptoPP::AsymmetricMultiply(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x18473a0 
  [4] CryptoPP::PositiveMultiply(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x1855167 
  [5] CryptoPP::Multiply(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x18551da 
  [6] CryptoPP::Integer::Times(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x18552c3 
  [7] CryptoPP::StringToInteger<char>(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x1862e12 
  [8] CryptoPP::Integer::Integer(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x184c9f5 
  [9] ValidateBBS(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x11b6dcb 
  [10] ValidateAll(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0x108bf07 
  [11] Validate(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0xcdb7fe 
  [12] main(0x0, 0x0, 0x0, 0x0, 0x0, 0x0), at 0xca2e7a 
(dbx) 
```

FloriansGit:
If no `CMAKE_BUILD_TYPE` is set, this does set the default to `RelWithDebInfo`. So it covers the intended functionality (e.g. setting `-g` which I think defaults to `-g2`) while being compatible with other build types/flags and platforms.

I also incorporated the changes suggested by Geoff in https://gist.github.com/geoffbeier/e9dcb0f623153ecdf042f14728eb878a, e.g. to check if the `CMAKE_CXX_FLAGS` are set or to print a message with the selected `CMAKE_BUILD_TYPE`. 

And I apologize for this being 4 commits in total (even I changed only one file and the author information of my last commit). I hope this not being a problem, I'm still not used to GitHub.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ set(cryptopp_VERSION_PATCH 5)
 
 include(GNUInstallDirs)
 include(TestBigEndian)
+include(CheckCXXCompilerFlag)
 
 #============================================================================
 # Settable options
@@ -112,9 +113,10 @@ if ((NOT CRYPTOPP_CROSS_COMPILE) AND (NOT (WINDOWS OR WINDOWS_STORE OR WINDOWS_P
 	endif()
 endif()
 
-# -march=native for GCC, Clang and ICC on i386 and x86_64.
-if ((NOT CRYPTOPP_CROSS_COMPILE) AND ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel"))
-	if (("${UNAME_MACHINE}" MATCHES "i.86") OR ("${UNAME_MACHINE}" STREQUAL "x86_64") OR ("${UNAME_MACHINE}" STREQUAL "i86pc"))
+# -march=native for GCC, Clang and ICC in any version that does support it.
+if ((NOT CRYPTOPP_CROSS_COMPILE) AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU|Intel"))
+	CHECK_CXX_COMPILER_FLAG("-march=native" COMPILER_OPT_ARCH_NATIVE_SUPPORTED)
+	if (COMPILER_OPT_ARCH_NATIVE_SUPPORTED AND NOT CMAKE_CXX_FLAGS MATCHES "-march=")
 		SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
 	endif()
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,17 @@ cmake_minimum_required(VERSION 2.8.5 FATAL_ERROR)
 
 project(cryptopp)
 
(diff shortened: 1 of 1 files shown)


noloader:
Thanks Forian. 

Some background... here's how we setup the library under Unix/Linux and Windows (using Linux as an example):

* Release build: `-DNDEBUG -g2 -O{2|3|5}`
* Debug build: `-DDEBUG -g3 -O0`
* Under-specified: `-DNDEBUG -g`

`-NDEBUG` must be set if the user *does not* specifically ask for a Debug build. We caught CVE-2016-7420 because other build systems, like Cmake and Eclipse, basically did what they wanted, and not what the project recommends.

The policy *"`-NDEBUG` must be set if the user *does not* specifically ask for a Debug build"* exists so users don't accidentally get into a debug configuration and then release to production. In a debug configuration, all the debugging and diagnostics are in effect, like asserts.

Asserts are the other half of our CVE-2016-7420. They can lead to an information disclosure if the assert fires. We use asserts liberally (let the program debug itself; I've got better things to do with my time), so there's a good chance an assert will fire.

-----

After all that, my question is, does the change meet the *"`-NDEBUG` must be set if the user *does not* specifically ask for a Debug build"* policy?

And a similar question is, does the build include `-g<something>` if the user does not specify otherwise. I've gotten two or three dumps and traces from Cmake users, and they are useless because there's no debug information.

noloader:
Thank you very much.

FloriansGit:
You're welcome. To answer your questions: this change is supposed to fix the common cases where the user didn't specify any build type at all. If a build type is specified you will get either `DEBUG` or `NDEBUG` defines and it will add debug information in the default case.

If you really want to enforce `DEBUG` or `NDEBUG` defines, I would recommend adding a test compile. Should I add a new issue for this?

noloader:
> If you really want to enforce DEBUG or NDEBUG defines, I would recommend adding a test compile. Should I add a new issue for this?

Any question that begins with *"Should we"* and ends with *"test"* gets a YES answer from me :) Feel free to fix it, add an issue and fix it, etc.

PR #339, opened 2016-12-01
Fix: GCC warning "type qualifiers ignored on function return type".

ralphtandetzky:
This pedantic message appeared all over the code. Also removed one warning about an unused variable in release build.


noloader:
I ran into this warning yesterday under GCC 6.2. Thanks for clearing them.

PR #341, opened 2016-12-02
Fix -Wshadow compiler warnings

droark:
When compiling using the -Wshadow compiler flag, quite a few warnings pop up. Clean up all instances of shadowing.

### DIFF
--- a/algebra.cpp
+++ b/algebra.cpp
@@ -205,9 +205,9 @@ template <class Element, class Iterator> Element GeneralCascadeMultiplication(co
 
 struct WindowSlider
 {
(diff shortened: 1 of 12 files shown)


noloader:
@droark,

Sorry about the late reply here. I seem to do that to you more than you deserve.

I think we should pass on this at the moment. I think prefixing parameters with `_` adds more diversity, and I think we should strive for consistency and not diversity. I think the way to proceed is to prefix the struct or class member with `m_`. The library seems to favor that style the most.

droark:
@noloader - No worries. I'll look into it when I get a chance. It could be awhile. For now, I'll close this out and reopen it later if necessary.

Thanks.

PR #351, opened 2016-12-27
spelling fixes

ka7:
patch contains some spelling-fixes, as found by a bot. ( https://github.com/ka7/misspell_fixer )
Should affect only text, /*comments*/ and other, non-critical stuff.
As there seems to be mixed win/*nix style, I see at least one ^M issue. Problem ?


noloader:
Thank you very much. All but two of those were mine. I'm an awful speller at times...

PR #352, opened 2016-12-28
Update Readme.txt

jaap198503:
Fixed typo.


noloader:
`cryptest` is actually correct. Also see [Visual Studio](https://www.cryptopp.com/wiki/Visual_Studio) on the Crypto++ wiki.

Issue #337, opened 2016-11-23
5.6.4 broke ABI of Integer::Integer()


wRAR:
The 5.6.4 release added a new argument of type CryptoPP::ByteOrder to several constructors of type CryptoPP::Integer. This broke the ABI of the class because adding an argument to a function, even with a defualt value, changes the function signature. See https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B for a list of what you can and what you cannot do to keep the ABI.

noloader:
Thanks @wRAR,

Yeah, it looks like the change had unwanted side effects. Below is from our [cryptest-symbols.sh](http://github.com/weidai11/cryptopp/blob/master/TestScripts/cryptest-symbols.sh) test script. We put it in place after we were bitten, but it was too late.

```
cp TestScripts/ cryptest-symbols.sh .
./cryptest-symbols.sh
...
cryptest.exe: symbol lookup error: /home/cryptopp/cryptest.exe: undefined symbol:
CryptoPP::Integer::Integer(unsigned char const*,unsigned long, CryptoPP::Integer::Signedness)
```

Also, see [Issue 283](http://github.com/weidai11/cryptopp/issues/283#issuecomment-258721959) and the ***Distros*** portion of the [Crypto++ 5.6.5 Release Notes](https://www.cryptopp.com/release565.html). There are at least two other subtle breaks that only distros experience.

My apologies for the troubles. As I said, we now have a process in place to actively hunt for these breaks rather than waiting for bug reports from user testing on a release candidate. These problems should not occur in the future.

Is there anything we can do to help you? Or is it OK to close this report?

wRAR:
Assuming the problem was not fixed yet, we have three options:

1. You fix the code so the symbol is there again (by adding a bunch of overloads).
2. You bump the soname.
3. We fix the code with a Debian-specific patch.

noloader:
@wRAR,

Email me off-list at *noloader, gmail account*.

This is the fourth or fifth subtle break for distros, so I want to put an end to it once and for all. For that, I want to email the distros and make sure everyone agrees with the path. I'll also loop-in a couple of folks from Crypto++ who help lead the project.

My gut instinct is to release 5.7. It won't have much in the way of new features, but it will clear these symbol problems. W can also get counsel on what to do with 5.6.x.

wRAR:
@noloader I am not the maintainer of the Debian package, you should contact him instead at gcs@debian.org

[...]

PR #356, opened 2017-01-04
Replace assert with void instruction

chfast:
In release builds replace assert with void instruction `(void)0`. Otherwise in some places you will end up with statements like `if (...) ;` and some compiler will complain about it.

### DIFF
--- a/trap.h
+++ b/trap.h
@@ -89,7 +89,7 @@
 // Remove CRYPTOPP_ASSERT in non-debug builds.
 //  Can't use CRYPTOPP_UNUSED due to circular dependency
 #ifndef CRYPTOPP_ASSERT
-#  define CRYPTOPP_ASSERT(exp)
+#  define CRYPTOPP_ASSERT(exp) (void)0
 #endif
 
 NAMESPACE_BEGIN(CryptoPP)


noloader:
>  Otherwise in some places you will end up with statements like `if (...) ;` and some compiler will complain about it.

Using the `if(ASSERT(....))` does not sound right. Its something we should fix.

Can you callout the problem areas?

chfast:
No. The code is `if (cond) ASSERT()`. In the end you will have "if" with
empty statement.

noloader:
Thank you very much. That looks like an unintended change that crept in with a documentation update. Sorry about that.

chfast:
Glad I could help. Do you plan any release soon?

noloader:
> Glad I could help. Do you plan any release soon?

Yes, I'd like to release soon.

The next release is going to be Crypto++ 5.7. We need the larger bump because we inadvertently broke the ABI, and caught issues like [337](https://github.com/weidai11/cryptopp/issues/337).

Because of the bump, I want to get a few (potentially) breaking changes that are not ready. For example, OCB and XTS modes. The effectively use a lightweight tweak, and they may need some not-so-readily apparent changes.

I also had a hard drive go bad in my test environment, so I lost about 600 GB of virtual machines. I need to build out  a handful of them for testing. For example, I need a new Debian 7, CentOS 5 and all the BSD machines because I know they cause problems on occasion.

Issue #355, opened 2017-01-04
CMake-based Xcode build fails to build any library


HeinrichJanzing:
I'm using CMake 3.6.1 and Xcode 8. I'm attempting to build cryptopp 5.6.5.

When attempting to build the static or shared library using cmake with the Xcode generator, the library itself is never created. The shared target fails due to the missing output file at some point while the static target claims to succeed despite the same lack of output. 

It appears to be caused by (at least for CMake >= 2.8.8) the shared and static library targets having only object files as input. According to [the documentation for cmake's add_library](https://cmake.org/cmake/help/latest/command/add_library.html#id4):

> Some native build systems may not like targets that have only object files, so consider adding at least one real source file to any target that references $<TARGET_OBJECTS:objlib>.

It would seem this is the case for Xcode. I see several options:
1. You tell me you don't care about Xcode (or any other affected generators), in which case I'll maintain a patch locally.
1. Applying the workaround suggested in the documentation. This could be done by creating an empty source file in the build directory and adding it to the target. A small and fairly local fix but hacky and perhaps even fragile.
1. The use of an intermediate objects target is made optional (rather than defined purely by the CMake version used) so it can be disabled manually in case of issues. This is my personal favorite. I would attach a patch for it but I'm having too much fun with the line endings. I'll try again tomorrow.

noloader:
Now open on the user list: [Issue 335: CMake-based Xcode build fails to build any library](https://groups.google.com/forum/#!topic/cryptopp-users/DzHhTBbRboQ)

mouse07410:
I don't think this report is correct. Here you can see a cmake build of the current Github master on MacOS Sierra, Xcode-8.2.1, Macports-installed cmake-3.7.1 (though it worked on previous versions of Mac OS X, older Xcode, and older cmake):
```
$ mkdir build
$ cd build
$ cmake ..
-- The C compiler identification is AppleClang 8.0.0.8000042
-- The CXX compiler identification is AppleClang 8.0.0.8000042
-- Check for working C compiler: /opt/local/bin/clang
-- Check for working C compiler: /opt/local/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /opt/local/bin/clang++
-- Check for working CXX compiler: /opt/local/bin/clang++ -- works
-- Detecting CXX compiler ABI info
[...]

PR #357, opened 2017-01-05
CMake: allow disabling the intermediate objects target (cryptopp-object).

HeinrichJanzing:
Targets with only object inputs do not work correctly with some
generators (like Xcode, see issue #355). Defining these directly in
terms of the source code files (rather than a reused set of object
files) allows correct builds in such cases. This can now be controlled
through a new option USE_INTERMEDIATE_OBJECTS_TARGET which defaults to
ON.


noloader:
@HeinrichJanzing,

Please forgive my ignorance. What targets are you having trouble with? The expected Crypto++ artifacts are `libcryptopp.a`, `libcryptopp.so` and `cryptest.exe`. I believe Cmake provides all three.

HeinrichJanzing:
They are indeed defined but in my experience, when using Xcode as the generator, they are never actually built, as described in the referenced issue #355. Both library targets fail to result in an actual library subsequently causing the cryptest.exe artifact to fail to build as well.

noloader:
> Both library targets fail to result in an actual library subsequently causing the cryptest.exe artifact to fail to build as well.

OK, thanks.

I think this is why Wei provided `adhoc.ccp.proto`. It predates Xcode, so I'm guessing it was added for Borland or Sun Studio or Metrowerks or some other old IDE.

What I am thinking... Can you add a recipe for Xcode that:

1. copies `adhoc.ccp.proto` to `adhoc.ccp`
2. adds the source file `adhoc.ccp` to the Xocde library targets?

My apologies for not doing it. I know very little about Cmake, and I really struggle with it.

noloader:
Merged. Thank you very much.

PR #359, opened 2017-01-08
Add dll export on some iterhash classes to fix CMake shared build.

egorpugin:


noloader:
This will likely break things for the static library when its being used to compliment the FIPS DLL. You can test the change by opening the Visual Studio solution, and then doing a Build → Batch Build → Build All. All 22 projects need to build correctly.

egorpugin:
Do you have some CI system like appveyor or travis?

egorpugin:
Yes, there're errors. That's why I asked if you ok with that PR.
It seems I won't be able to resolve those errors, because I'm not familiar with provided sln build structure.

egorpugin:
What about the following?
```
#ifndef CRYPTOPP_ENABLE_COMPLIANCE_WITH_FIPS_140_2
#define CRYPTOPP_DLL_NO_FIPS CRYPTOPP_DLL
#else
#define CRYPTOPP_DLL_NO_FIPS 
#endif

class CRYPTOPP_DLL_NO_FIPS CRYPTOPP_NO_VTABLE IteratedHashBase : public BASE
```

noloader:
@egorpugin, I'm going to close this PR. I'll add some documentation on how to build a *Regular DLL* on Windows.

PR #363, opened 2017-01-13
Suppress warning also for VS2017

MarcelRaad:
The warning suppression for Visual Studio 2015 is also needed for Visual Studio 2017 (_MSC_VER = 1910).


noloader:
Thanks.

Man, they still have not fixed that?

Issue #364, opened 2017-01-15
Apple Clang 6.0 and numeric_limits<word128>::max() returns 0


noloader:
Testing on OS X 10.9 using Apple Clang 6.0. `cryptest.sh` reveals `TestRounding()` is failing one test case:

```
try
{
    const word128 h = ((word128)W64LIT(0xffffffffffffffff)) << 64U;
    const word128 v = h | (word128)W64LIT(0xfffffffffffffff9), b=0x08;
    word128 r=RoundUpToMultipleOf(v, b);
    fail = true;
}
catch(const Exception&)
{
    fail = false;
}
```

Here's the debug session with `const word128 m = std::numeric_limits<word128>::max();` added for instrumentation.

```
Breakpoint 1, TestRounding () at validat0.cpp:456
456                     const word128 h = ((word128)W64LIT(0xffffffffffffffff)) << 64U;
(gdb) n
457                     const word128 v = h | (word128)W64LIT(0xfffffffffffffff9), b=0x08;
(gdb)
458                     const word128 m = std::numeric_limits<word128>::max();
(gdb)
459                     word128 r=RoundUpToMultipleOf(v, b);
(gdb) p v
$1 = 0xfffffffffffffffffffffffffffffff9
(gdb) p m
$2 = 0x00000000000000000000000000000000
```

The tests were put in place after [Issue 360](https://github.com/weidai11/cryptopp/issues/360).

anonimal:
>Testing on OS X 10.9

10.9 reached end of life last year. Will cryptopp continue 10.9 support? Does cryptopp have any 10.10+ machine(s) available to do testing on?

noloader:
> 10.9 reached end of life last year. Will cryptopp continue 10.9 support?

Yes, we support as far back as we reasonably can. OS X is a great example of why. Apple abandoned the platform, so the only way for users to stay up to date is to use up-to-date libraries.

Projects like Brew, Fink and MacPorts are a great help. But they depend on projects like Botan, Crypto++ and OpenSSL maintaining libraries for them to use.

> Does cryptopp have any 10.10+ machine(s) available to do testing on?

I do not. I have OS X 10.5, 10.8 and 10.9. I believe Uri tests on OS X 10.10 or 10.11.

We also have gaps in testing on iPhones and iPads. I have 5 iPads, 1 iPhone and 1 iPod. All of them are older and abandoned by Apple. I think the latest OS I can test is iOS 8 or iOS 9.

[...]

PR #368, opened 2017-01-24
Support for AES-256-CBC with W3C padding scheme ( http://www.w3.org/2001/04/xmlenc#aes256-cbc )

danielweck:
Quoted from:
https://github.com/readium/readium-lcp-client/pull/26#issuecomment-274689831

"
IMO, we need to consider why W3C suddenly added the padding restriction in 2002.

There was no padding details before 2002 in the W3C encryption standard. But after publishing the possibility of the Padding Oracle Attack on the block cipher algorithm in 2002, W3C had instantly updated the specification as size of the padding bytes should not be described in the padding block except last byte.

Key point of the padding oracle is that the last encrypted block before the PKCS#7 padding could be easily cracked, if the decryption program provides unnecessary error codes related padding.

So I think that not allowing using the vulnerable PCKS#7 on both server and client side is the clear way for not only consistency of the specification, but also the security perspective.
"

CC @thkim2015 @drminside

### DIFF
--- a/filters.cpp
+++ b/filters.cpp
@@ -701,6 +701,7 @@ void StreamTransformationFilter::LastPut(const byte *inString, size_t length)
 
 	case PKCS_PADDING:
 	case ONE_AND_ZEROS_PADDING:
+	case W3C_PADDING:
 		unsigned int s;
 		s = m_cipher.MandatoryBlockSize();
 		CRYPTOPP_ASSERT(s > 1);
@@ -734,6 +735,13 @@ void StreamTransformationFilter::LastPut(const byte *inString, size_t length)
 				if (pad < 1 || pad > s || std::find_if(space+s-pad, space+s, std::bind2nd(std::not_equal_to<byte>(), pad)) != space+s)
 					throw InvalidCiphertext("StreamTransformationFilter: invalid PKCS #7 block padding found");
 				length = s-pad;
+			}
+			else if (m_padding == W3C_PADDING)
+			{
+				byte pad = space[s - 1];
+				if (pad < 1 || pad > s)
+					throw InvalidCiphertext("StreamTransformationFilter: invalid W3C block padding found");
+				length = s - pad;
 			}
 			else
 			{
--- a/filters.h
+++ b/filters.h
@@ -480,6 +480,8 @@ struct BlockPaddingSchemeDef
 		PKCS_PADDING,
 		//! \brief 1 and 0's padding added to a block
 		ONE_AND_ZEROS_PADDING,
+		//! \brief [Random bytes (0 ~ N-2) and padding's length (N-1)]'s padding to a block
+		W3C_PADDING,
 		//! \brief Default padding scheme
 		DEFAULT_PADDING
 	};


noloader:
Thanks @danielweck.

The PR message was posted at [PR 368: Support for AES-256-CBC with W3C padding scheme](https://groups.google.com/forum/#!topic/cryptopp-users/72Kzsp8QOe0). It is a process formality.

Would you happen to know if test vectors exist? I'd like to add them independent of this patch. That is, I'll cut them in, and they don't need to be part of the patch.

noloader:
Sent to the W3C editors and authors.

danielweck:
@noloader thanks for your prompt feedback!
Let me clarify that the original authors for this CryptoPP patch are @thkim2015 / @drminside (all credits to them).
I thought I'd publish this PR as I am in the process of re-organising / cleaning-up the Crypto++ fork that's used in the Readium LCP project. Besides support for AES-256-CBC W3C padding scheme, we have minor XCode etc. project additions. Nothing else worth contributing back.
Cheers!
Dan

danielweck:
PS: also note that resources encrypted with AES-256-CBC PKCS#7 padding scheme can successfully be decrypted using the W3C-padding implementation proposed in this PR.

noloader:
@danielweck,

> PS: also note that resources encrypted with AES-256-CBC PKCS#7 padding scheme can successfully be decrypted using the W3C-padding implementation proposed in this PR (but not vice-versa)

Ack thanks.

> ... have minor XCode etc. project additions

Let me know if we can help. As long as Xcode changes don't break other builds, we can probably incorporate them.

We test Apple operating systems like OS X and iOS, but WatchOS and AppleTV have gaps. I don't think they are being tested at the moment. Also see [Issue 364](https://github.com/weidai11/cryptopp/issues/364#issuecomment-272723127).

danielweck:
Additional project files:
https://github.com/edrlab/cryptopp/compare/CRYPTOPP_5_6_5...CRYPTOPP_5_6_5_XCode_etc_projects

noloader:
@danielweck, Did test vectors materialize? Or, is there another (independent) implementation around that I can generate them from?

danielweck:
no exiting news from my end I'm afraid.

noloader:
@danielweck,

I think the patch needs a small addition. It looks like the decryption path is has been added, bit it appears the encryption path is missing. Below is the encryption path. Because it is ***not*** the case `m_padding == PKCS_PADDING`, you will fall into the `else`, which is `ONE_AND_ZEROS_PADDING`.

```
if (m_cipher.IsForwardTransformation())
{
    CRYPTOPP_ASSERT(length < s);
    if (inString) {memcpy(space, inString, length);}
    if (m_padding == PKCS_PADDING)
    {
        CRYPTOPP_ASSERT(s < 256);
        byte pad = byte(s-length);
        memset(space+length, pad, s-length);
    }
    else
    {
        space[length] = 0x80;
        memset(space+length+1, 0, s-length-1);
    }
    m_cipher.ProcessData(space, space, s);
    AttachedTransformation()->Put(space, s);
}
else
{
    // decryption
}
```

Though the W3C scheme says the fill bytes are arbitrary, it looks like it could lead to an information disclosure if not careful. For example, suppose B=16 and m={1,2,3,4} (and X= "don't care"). The W3C padded string would be {1,2,3,4, X,X,X,X, X,X,X,X, X,X,X,12}. If the 11 each "don't cares" are left as is, then the other party gets to view that memory.

I would feel better if the "don't care" bytes were memset'd to 0 or use PKCS5 padding to avoid the potential disclosure. I think the other option is to use random values, but Crypto++ is not really setup for it in this area.

What do you think?

thkim2015:
@noloader 

Could you explain what you exactly mean the other party gets to view the arbitrary part if it is not set to specific value?
And I am sure what kind of risks there are, if it is disclosed as it is.

Although meaning 'Arbitrary' includes PKSCS#5(7) and always-specific value such as 0, however if the decryption module is not carefully operated on the padding information, it says some hacking program could get the decryption value without knowing key. It means the decryption module would be better not to know the values of the arbitrary part. So I think it could be the more risky if the decryption module has to check the arbitrary part should be specific value.

noloader:
@danielweck,

I'll be merging this today. There were no complaints from the folks on the user list.

We will need to tweak it by adding a path to add padding on the encryption path. The following test program:

```
int main(int argc, char* argv[])
{
	byte key[16] = {0}, iv[16] = {0};

	string str = "Hello World";
	string encrypted, recovered;
	
	CBC_Mode<AES>::Encryption enc(key, 16, iv);
	StringSource ss1(str, true, new StreamTransformationFilter(enc,
			new StringSink(encrypted), BlockPaddingSchemeDef::W3C_PADDING));
			
	CBC_Mode<AES>::Decryption dec(key, 16, iv);
	StringSource ss2(encrypted, true, new StreamTransformationFilter(dec,
			new StringSink(recovered), BlockPaddingSchemeDef::W3C_PADDING));
	
	cout << "Recovered: " << recovered << endl;

    return 0;
}
```

Causes an exception:

```
$ ./test.exe
terminate called after throwing an instance of 'CryptoPP::InvalidCiphertext'
  what():  StreamTransformationFilter: invalid W3C block padding found
Aborted (core dumped)
```

I believe its due to `ONES_AND_ZEROS_PADDING`.

Thanks again.

noloader:
Also see [Commit 8e088b48654f78cb](https://github.com/weidai11/cryptopp/commit/8e088b48654f78cb9bc0643276e1b7e1168e0571)

danielweck:
Awesome additions @noloader ! :)

Credits to @thkim2015 @drminside for the original code in this pull request, and thanks to @edrlab for hosting the actual fork :)   (which I will update as soon as the next official CryptoPP release comes out)

By the way, any idea when version `5.6.6` is likely to hit the release area?
https://github.com/weidai11/cryptopp/releases

noloader:
> By the way, any idea when version 5.6.6 is likely to hit the release area?

The next release will be 5.7 or 6.0. We inadvertently broke the ABI at 5.6.5, so we need a minor version bump to 5.7. We want to tweak the KDF interface so we can add modern derivation functions, and that will need a major version bump.

I don't know when it will be ready. We hope it will be in the next couple of months.

You might also consider pinging Jack Lloyd. He's the author of [Botan](https://github.com/randombit/botan). Its another good choice for C++ security libraries. Increasingly, I am using it for cross-validation.

danielweck:
Thanks for the pointers!

noloader:
Also see [Commit 8a177c58e692](https://github.com/weidai11/cryptopp/commit/8a177c58e692f8852199868ff4845deac55b461b).

Issue #391, opened 2017-03-14
Crash in NonblockingRng::GenerateBlock


janblome:
On windows `NonblockingRng::GenerateBlock` initializes a static reference to the `MicrosoftCryptoProvider` (see [osrng.cpp](/weidai11/cryptopp/blob/CRYPTOPP_5_6_5/osrng.cpp#L153)). This can lead to a crash when the code was compiled with Visual Studio 2013 because local static variables are not thread-safe, as mentioned in #372.

noloader:
Thanks @janblome.

It looks like we are going to have to forgo C++11 synchronization gear on Microsoft platforms.

noloader:
OK, so I *think* this is could be a problem with [Dynamic Initialization and Destruction with Concurrency](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2660.htm). Dynamic Initialization and Destruction with Concurrency is a core language feature set forth in 2008. I can't imagine Microsoft implementations lack it considering its almost a decade old, Clang provided it in 2.8, and GCC provided in in 4.3. Additionally, [CPP Reference | C++ 2011 features ](http://en.cppreference.com/w/cpp/compiler_support) states Microsoft compilers support it since `CL` 14.00/VS 2008.

Here's what I am thinking for `config.h`. It adds a define for `CRYPTOPP_MSC_VERSION` and basically disables it for almost all Microsoft compilers and some platforms. Until we figure out what ***does not*** crash under MS toolchains, programs written for Windows will just use the old C++03 Singleton. The old Singleton could leak memory, but it does not crash and burn.

```
// Dynamic Initialization and Destruction with Concurrency
// MS at VS2018 (20.00); GCC at 4.3; Clang at 2.9; Intel 11.1; SunCC 5.13.
#if (CRYPTOPP_MSC_VERSION >= 2000) && ((WINVER >= 0x0600 /*_WIN32_WINNT_VISTA*/) || (_WIN32_WINNT >= 0x0600 /*_WIN32_WINNT_VISTA*/))
#  define CRYPTOPP_CXX11_DYNAMIC_INIT 1
#elif (CRYPTOPP_LLVM_CLANG_VERSION >= 30300) || (CRYPTOPP_APPLE_CLANG_VERSION >= 50000)
#  define CRYPTOPP_CXX11_DYNAMIC_INIT 1
#elif (__INTEL_COMPILER >= 1110)
[...]

PR #392, opened 2017-03-15
Remove unnecessary static from reference to singleton

janblome:
While 46c9cc725cf325f2 fixed the static initialization problem with Visual Studio in the singleton, there is still a race condition in `NonblockingRng::GenerateBlock` (see issue #391). The reference to the singleton is still a function-level static and it will still crash if a thread enters the function while another thread is already initializing the variable. To fix this I simply removed the static from the variable.


noloader:
Thank you very much.

PR #394, opened 2017-03-20
Improve sha256 performance on ppc64 by 4.5x

gut:
Results: I removed the other hash algorithm on test.cpp and gave as an
input a ~700MB file.

Upstream:
$ time ./cryptest.exe m ~/ubuntu-16.10-server-ppc64el.iso
SHA-256: d14bdb413ea6cdc8d9354fcbc37a834b7de0c23f992deb0c6764d0fd5d65408e

real    0m18.811s
user    0m18.456s
sys     0m0.356s

This patch:
$ time ./cryptest.exe m ~/ubuntu-16.10-server-ppc64el.iso
SHA-256: d14bdb413ea6cdc8d9354fcbc37a834b7de0c23f992deb0c6764d0fd5d65408e

real    0m4.158s
user    0m3.992s
sys     0m0.168s

This approach used altivec + VSX instructions found on POWER8 systems
and newer.

If unwanted on PPC, "cmake" it with -DDISABLE_ALTIVEC=1 or "make" it
with USE_ALTIVEC=0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ endif()
 option(DISABLE_ASM "Disable ASM" OFF)
 option(DISABLE_SSSE3 "Disable SSSE3" OFF)
 option(DISABLE_AESNI "Disable AES-NI" OFF)
+option(DISABLE_ALTIVEC "Disable PPC64's Altivec registers for SHA-2" OFF)
 option(DISABLE_NATIVE_ARCH "Disable the addition of -march=native" OFF)
 set(CRYPTOPP_DATA_DIR "" CACHE PATH "Crypto++ test data directory")
 
@@ -96,6 +97,9 @@ if(DISABLE_SSSE3)
 endif()
 if(DISABLE_AESNI)
 	add_definitions(-DCRYPTOPP_DISABLE_AESNI)
+endif()
+if(DISABLE_ALTIVEC)
+  add_definitions(-DCRYPTOPP_DISABLE_ALTIVEC)
 endif()
 if(NOT CRYPTOPP_DATA_DIR STREQUAL "")
 	add_definitions(-DCRYPTOPP_DATA_DIR="${CRYPTOPP_DATA_DIR}")
@@ -240,6 +244,15 @@ if(MSVC AND NOT DISABLE_ASM)
 		list(APPEND cryptopp_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/rdrand.asm)
 		enable_language(ASM_MASM)
 	endif()
+endif()
+
+if (${UNAME_MACHINE} MATCHES "ppc64le" AND NOT ${DISABLE_ALTIVEC})
+	message(STATUS "Enabling SHA-2 acceleration for ppc64le.")
+	set(cryptopp_SOURCES
+			${CMAKE_CURRENT_SOURCE_DIR}/sha256_compress_ppc.s
+			${cryptopp_SOURCES}
+			)
+	enable_language(ASM)
 endif()
 
 #============================================================================
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -57,6 +57,9 @@ IS_X32 ?= 0
 # Set to 1 if you used NASM to build rdrand-{x86|x32|x64}
 USE_NASM ?= 0
 
+# Set to 0 if you don't want acceleration on SHA on PPC64le
+USE_ALTIVEC ?= $(IS_PPC)
+
 # Fixup for X32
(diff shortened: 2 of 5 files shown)


noloader:
Ack, thanks @gut. Give me a couple of days for the cut-in.

noloader:
@gut,

We got out first Power8 implementation behind us; see [Add Power8 AES Encryption](https://github.com/weidai11/cryptopp/issues/497). AES performance is very impressive. Its about 20x faster than C/C++.

We'd like to pursue SHA now. Do you have some spare cycles to work with us?

gut:
Hi!
Great! Huge boost!
What do you have in mind? The whole SHA-2? This pull request added only SHA-256.

noloader:
@gut,

Thanks for the continued interest, and sorry about the late reply. We are in a much better position to support you now.

So we started tracking the SHA addition at [Issue 513, Add Power8 SHA Hashing](https://github.com/weidai11/cryptopp/issues/513). Its got several supporting commits based on our experience with the [AES cut-in](https://github.com/weidai11/cryptopp/issues/497).

The first two commits are the library abstracting Linux, AIX, GCC and XLC and hiding the differences behind a common interface in a single header. Effectively we cleaned up the mess you would encounter in C/C++. It also handles that cursed `vec_sld` (`vsldoi`) that breaks on little endian.

The second is, the header now has `VectorSHA256` and `VectorSHA512` for you. `VectorSHA256` and `VectorSHA512` are template functions due to the "compile time constant" requirement. You would call them like so, depending on whether you wanted `sigma0`, `sigma1`, `Sigma0` or `Sigma1`.

```
uint8x16_p8 x;
x = VectorSHA256<0,0>(x);
```

The third is, we stubbed-out SHA for you so you don't need to hijack C/C++. You need to open `config.h` and uncomment `CRYPTOPP_POWER8_SHA_AVAILABLE`. Then, work your magic in `sha-simd.cpp`.

gut:
Hi @noloader 
I can't really tell right now if my company allows to license it under public domain, but probably not. I'll get that back to you soon.

Either way, our SIMD implementation is available, at least for reference purposes, on https://github.com/PPC64/sha2-le so you can feel free to use that.

Also I don't know if I'll be able to help porting the SHA-512 as my project priorities shifted a bit and I'm not looking into theses things at the moment. So I'll be up to you, ok?

Thanks for getting back!

noloader:
Closing again. We still did not get to it. It will be on the top of the TODO list.

noloader:
@gut,

We finally rolled an implementation using Power8 SHA intrinsics. See my incubator at [Noloader | SHA-Intrinsics](https://github.com/noloader/SHA-Intrinsics).

The performance absolutely sucks. I think your implementation probably outperforms the SHA intrinsic implementation. IBM should be ashamed of themselves for such an under-performing waste of time.

gut:
It does sucks, but it also may be that we used it incorrectly (think about inline assembly barriers and how the compiler is being extra careful when handling that code), but as it was a long time ago, I really don't remember how far we made it into this investigation (mostly we used disassembly to find out why that code was inefficient).

In the end we opted to do the whole assembly by ourselves, and it's now precompiled by the gnu's m4 preprocessor to match some architecture configuration (like endianness and alignment of input data):
https://github.com/PPC64/sha2-le/blob/master/sha256_compress_ppc.m4#L4-L12

noloader:
@gut,

> It does sucks, but it also may be that we used it incorrectly...

Yeah, I agree with you. The problem is IBM did not tell anyone how to use it. The company did not provide a whitepaper, and the company  did not provide a reference implementation. Even Andy's implementation for OpenSSL underperforms.

The onus is on IBM to provide the necessary documentation.

noloader:
Thanks again @gut.

We checked-in an intrinsic/built-in based implementation at [Commit 0630d46fe828](https://github.com/weidai11/cryptopp/commit/0630d46fe828).

We know the little-endian numbers are off because the compiler is screwing up the loads. Also see [GCC Issue 84753 - GCC does not fold xxswapd followed by vperm](http://gcc.gnu.org/bugzilla/show_bug.cgi?id=84753). What is happening is, the compiler is generating `lvx` followed by `xxswapd` and `xxlnand` to correct endianess, and _then_ it is applying our `vperm` mask. So there are four instructions to load and permute an arrangement.

The big-endian numbers look a lot better but they probably lag behind you implementation. Big-endian does not suffer the little-endian brain dead-ness. It looks like there's some inefficiency in the way registers are allocated and offloaded. I don't know how to control it from C/C++ code.

gut:
@lbianc and @leitao : ping. Please take a closer look to this.

Issue #385, opened 2017-03-01
Don't use WSAStringToAddress, use WSAStringToAddressA instead


asbai:
Otherwise it can't compiled under UNICODE mode. Because the code always use an ANSI char to receive the result:
```
	char temp[MAX_ADDRSTRLEN]; // ANSI char
	// ...
	if (WSAStringToAddress(temp, af, NULL, (struct sockaddr *)&ss, &size) == 0) { // ...
```
`WSAStringToAddress` is mapping to `WSAStringToAddressW` under UNICODE version, and `WSAStringToAddressA` under the ANSI version. Explicitly using `WSAStringToAddressA` if you always want to use the ANSI one.

noloader:
Thanks @asbai.

It sounds like the build system is not quite correct. Can you verify [`CharacterSet` is set to `MultiByte` in `cryptlib.vcxproj`](https://github.com/weidai11/cryptopp/blob/master/cryptlib.vcxproj#L56):

```
  <!-- All Configurations -->
  <PropertyGroup Label="All Configurations">
    <ConfigurationType>StaticLibrary</ConfigurationType>
    <TargetName>cryptlib</TargetName>
    <TargetExt>.lib</TargetExt>
    <UseOfMfc>false</UseOfMfc>
    <CharacterSet>MultiByte</CharacterSet>
    <WholeProgramOptimization>true</WholeProgramOptimization>
</PropertyGroup>
```

asbai:
Ha, I'm not using the default vcproj and solution file belong with the source package. Actually, I'm build crypto++ by my own vc project files before many years ago. There is many reason to do that, for example:  I need many components from different libraries can interactive with each others smoothly.

But I think it is not a "configuration problem", it is a code problem. Because we are using the MSBC (ANSI) char with a "TCHAR" API (`WSAStringToAddress` talk TCHAR). If we really want to talking with char, we should use the ANSI version (`WSAStringToAddressA`) to ensure it. So we can talk with MBCS charset no matter what configuration are using.

denisbider:
@asbai is right that it is generally a good idea to identify the Windows function version (ANSI/Unicode) that the library wants to call explicitly, instead of relying on the macro. The only reason one would not identify the function version (A or W) is if you're writing code that uses TCHAR and is meant to be compiled as either ANSI or Unicode. However, this code clearly needs the A version of the function, so it's a good idea to specify the A version instead of using the A/W macro.

denisbider:
[...]

PR #397, opened 2017-03-26
Fix Windows compilation issues

RaptorFactor:
This fixes compilation under:

- Any compiler on Windows with /DUNICODE
- MSVC with /std:c++latest
- MSVC targeting ARM
- Clang-cl (i.e. Clang pretending to be MSVC) targeting x86/x64


RaptorFactor:
Which parts will break C++03? I can look to see if I can fix that.

Also all the tests pass for me when targeting clang-cl. The declspec(novtable) problem is already worked around in this patch, and that was the only runtime error when running the tests.

MarcelRaad:
> Also all the tests pass for me when targeting clang-cl. The declspec(novtable) problem is already worked around in this patch, and that was the only runtime error when running the tests.

Works for me too, there's no crash anymore 👍 

> The workaround I implemented to remove std::bind2nd should work in C++03, because both std::bind and std::placeholders::_1 are present in C++03.

Unfortunately only in TR1, and in its std::tr1 namespace.

RaptorFactor:
Doh you're right. Been too long since I've had to use C++03 (though to be fair C++11 has been widely available for half a decade).

I'm not sure of a good workaround in that case.... I could do a version check on __cplusplus if that's acceptable and just provide both implementations?

noloader:
I'm willing to meet you half way:

1. Leave the "unsupported" in place. Users who want it can experiment.
2. Leave the `std::bind2nd` in place. I test under 6 compilers from C++03 through C++17 and it has never caused a problem.
3. Don't use `std::placeholder` since its not C++03.
4. Take the `RDRAND`, `RDSEED`, `Integer`, `_M_ARM` and `OutputDebugString` changes.

The reason  for (1) is I had to take on CMake maintenance because the folks who wanted it mostly lost interest after we added `CMakeList.txt`. I don't want Clang on Windows maintenance because I know how difficult Clang is to work with at times.

I'm not sure about `__crt_debugger_hook`. I test compiles under ARM with `cryptest.nmake` and I have never encountered a problem. What is the issue?

The `OutputDebugString` changes were on the table a while ago. No one was interested in the change at the time, so things were left as-is. Also see [Issue 385, Don't use WSAStringToAddress, use WSAStringToAddressA instead](https://github.com/weidai11/cryptopp/issues/385#issuecomment-286436615).

If interested in maintaining LLVM on Windows support, you should set up some VMs with Vista through Windows 10 and VS2008 through VS2015 and test it more thoroughly. The earliest LLVM on Windows download is Clang 3.4. Some problems are just unsolvable, like that standard header problem.

MarcelRaad:
@noloader @RaptorFactor 
> 1.Leave the "unsupported" in place. Users who want it can experiment.

Maybe something like this would be OK so that we can just define something in the project file instead of patching the library after every update?

```
// Clang pretends to be VC++, too.  
//   See http://github.com/weidai11/cryptopp/issues/147  
#if defined(_MSC_VER) && defined(__clang__) && !defined(CRYPTOPP_ENABLE_UNSUPPORTED_MSVC_CLANG)
# error: "Unsupported configuration"  
#endif  
```

noloader:
Does anyone know where we are with this PR?

I see `std::bind2nd` is still missing, and using `std::bind` and `std::placeholders` instead. If I am parsing things correctly, it looks like the original PR.

MarcelRaad:
Yes, the "Fix compilation with MSVC and /std:c++latest" commit still breaks C++03. Could you please remove it from this PR @RaptorFactor and maybe make another PR for that issue replacing them conditionally on C++11 support? Probably a new feature macro for std::bind support in config.h would make sense.

Or you could just cherry-pick the other commits from refs/pull/397/head @noloader.

noloader:
Marcel provided a PR that removed the things that would break C++03. Also see [Pull Request 444](https://github.com/weidai11/cryptopp/pull/444).

We merged the updated PR today. Closing this PR.

Thanks for everyone's help with it.

Issue #395, opened 2017-03-20
OpenBSD (egcc/eg++): Assembler error: no such instruction


anonimal:
System info:
- OpenBSD 6.0 (GENERIC)
- cpu0: Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz, 3500.59 MHz
- cpu0: FPU,VME,DE,PSE,TSC,MSR,PAE,MCE,CX8,APIC,SEP,MTRR,PGE,MCA,CMOV,PAT,PSE36,CFLUSH,MMX,FXSR,SSE,SSE2,SSE3,PCLMUL,MWAIT,SSSE3,CX16,SSE4.1,SSE4.2,POPCNT,AES,XSAVE,AVX,RDRAND,NXE,LONG,LAHF,ITSC

Notes:
- Only producible with egcc/eg++ 4.9.3 (no errors with other tested compilers: clang 3.8 or gcc 4.2.1)
- `-DDISABLE_ASM=ON` reduced number of errors but the issue still remains
- **`-DDISABLE_NATIVE_ARCH=ON` patches the issue** (I imagine env exporting from cli will also work)

Build:
- Built against f502ee9218c3a090920bc1584334a3a8e3fbb043
- [Cryptopp-f502ee9-OpenBSD-build-error.CMake.log.txt](https://github.com/weidai11/cryptopp/files/856184/Cryptopp-f502ee9-OpenBSD-build-error.CMake.log.txt)
- [cryptest-result.txt](https://github.com/weidai11/cryptopp/files/856211/cryptest-result.txt)

noloader:
Thanks @anonimal. And sorry about that.

Let me get a VM fired up for it. I think I have one somewhere.

noloader:
@anonimal,

The case I'm not sure about is tying `DISABLE_NATIVE_ARCH` to `CRYPTOPP_DISABLE_ASM`. `CRYPTOPP_DISABLE_ASM` is intended to be the fail safe if the library's ASM is not suitable for a platform.  Similar options, like `CRYPTOPP_DISABLE_SSSE3`, are a more surgical application of `CRYPTOPP_DISABLE_ASM` for old assemblers. Another use case we are now seeing is "new" compilers with "old" platform components like GAS 2.17, and that begs `DISABLE_NATIVE_ARCH` option.

[...]

PR #398, opened 2017-03-29
Build: apply patch for GNU assembler errors to CMake

anonimal:
References weidai11/cryptopp#395

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,57 @@ if(NOT CRYPTOPP_DATA_DIR STREQUAL "")
 	add_definitions(-DCRYPTOPP_DATA_DIR="${CRYPTOPP_DATA_DIR}")
 endif()
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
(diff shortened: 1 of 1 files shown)


anonimal:
I'm not a huge fan of CMake, so I'm sure there are better ways to go about doing this.

noloader:
@anonimal,

Have a quick look at this question on Stack Overflow: [Determine Assembler version in CMake?](http://stackoverflow.com/q/43059015/608639)

anonimal:
@noloader Ah, very nice. Straightforward and simple. Thanks for the link! I'll rebase sometime within the next 24 hours.

anonimal:
(It would be nicer if it worked. I'll work out the kinks and see what I can do.)

anonimal:
@noloader I'm running into regex problems similar to my TODO in the commit (yes, slashes are properly escaped unlike the example response in that SO link). Other than that, the identifier works fine.

What I've PR'd works but it can obviously be improved. I'm not keen on doing any more work on this issue but I also don't want to leave this PR in limbo if you're not happy with it.

How would you like me to pursue a resolution?

noloader:
Thank you very much.

Issue #403, opened 2017-04-22
Solaris 10 and SIGBUS crashes in cryptest.exe


noloader:
See [Coredump in cryptest.exe on Solaris 10](https://groups.google.com/d/msg/cryptopp-users/OYaByDEbSI0/624NTaBmBQAJ) on the mailing list and [Pull Request 402](https://github.com/weidai11/cryptopp/pull/402). Multiple self tests fail when `data` is read as a `word64`. Alignment appears to be OK.

The unusual thing is, `memcpy`-out the `data` parameter to a `word64` clears some of the issues. But the debugger shows the `data` parameter is aligned (from the core dump):

```
Testing MessageDigest algorithm SHA-384.
..signal BUS (invalid address alignment) in CryptoPP::SHA512::Transform at line 1762 in file "sha.cpp"
 1762           R( 0); R( 1); R( 2); R( 3);^M
(dbx) where
=>[1] CryptoPP::SHA512::Transform(state = 0x101167a08, data = 0xffffffff7fffa68c), line 1762 in "sha.cpp"
  [2] CryptoPP::SHA384::Transform(digest = 0x101167a08, data = 0xffffffff7fffa68c), line 85 in "sha.h"
  [3] CryptoPP::IteratedHashWithStaticTransform<unsigned long,CryptoPP::EnumToType<CryptoPP::ByteOrder,1>,128U,64U,CryptoPP::SHA384,48U,0>::HashEndianCorrectedBlock(this = 0x101167950, data = 0xffffffff7fffa68c), line 165 in "iterhash.h"
  [4] CryptoPP::IteratedHashBase<unsigned long,CryptoPP::HashTransformation>::HashMultipleBlocks(this = 0x101167950, input = 0xffffffff7fffa68c, length = 2852U), line 94 in "iterhash.cpp"
...
```

noloader:
Merged PR.

kvirund:
> But the debugger shows the data parameter is aligned (from the core dump)

`data` is aligned to 4, but not aligned to 8 bytes:

```
>>> print(0xffffffff7fffa68c % 8)
4
>>> print(0xffffffff7fffa68c % 4)
0
>>> 
```

noloader:
Arg... I'm not sure why I made the leap that an address that ends in 0xc is aligned for a 64-bit type on SPARC.

[...]

PR #402, opened 2017-04-21
Some fixes related to SunCC compiler bugs.

kvirund:
- Fixes for "invalid address alignment" errors. See https://groups.google.com/d/topic/cryptopp-users/OYaByDEbSI0/discussion for details.
- Changed config.h in accordance to Jeffrey Walton's request;

There is still error with

```
DH validation suite running...


signal BUS (invalid address alignment) in CryptoPP::BufferedTransformation::GetMaxWaitObjectCount at line 399 in file "cryptlib.cpp"
  399           return t ? t->GetMaxWaitObjectCount() : 0;^M
```

PR #405, opened 2017-04-27
Added -xmemalign=4i for Sparc

kvirund:
Also reverted previously committed workarounds for this issue (see #402, #403).


noloader:
Thank you very much.

Issue #415, opened 2017-05-11
BLAKE2 produces wrong hashes for non-default digest sizes


c0ff:
The constructor `BLAKE2_Base(bool treeMode, unsigned int digestSize)` uses an incorrect initialization block and produces wrong results if either treeMode or digestSize differ from default.
The patch is here:
[blake2-init-patch.zip](https://github.com/weidai11/cryptopp/files/991589/blake2-init-patch.zip)

In essense it replaces
`UncheckedSetKey(NULL, 0, g_nullNameValuePairs);`
with
`UncheckedSetKey(NULL, 0, MakeParameters(Name::DigestSize(), (int)digestSize)(Name::TreeMode(), treeMode, false));`

noloader:
Thanks @c0ff.

Can you test this against master? I'm tied up at the moment with other testing.

```
git clone https://github.com/weidai11/cryptopp-fix
cd cryptopp-fix

./cryptest.sh
```

`cryptest.sh` is our test script from hell, and it torture tests the library under different configurations. On a modern desktop, like a 5th gen or 6th gen, it will take a couple of hours to run.

c0ff:
Adjusted and tested the patches against master: [blake2-init-and-test-patches.zip](https://github.com/weidai11/cryptopp/files/991767/blake2-init-and-test-patches.zip)

Though, I only did a short test with `cryptest.exe v 74` and `cryptest.exe v 75`.

After applying blake2-test.patch:
  Without blake2-init.patch:
    [failed_blake2s.txt](https://github.com/weidai11/cryptopp/files/991768/failed_blake2s.txt)
    [failed_blake2b.txt](https://github.com/weidai11/cryptopp/files/991770/failed_blake2b.txt)

  With blake2-init.patch:
    [passed_blake2s.txt](https://github.com/weidai11/cryptopp/files/991773/passed_blake2s.txt)
    [passed_blake2b.txt](https://github.com/weidai11/cryptopp/files/991774/passed_blake2b.txt)

c0ff:
The new test vectors were generated with pyblake2 which uses 'BLAKE2 reference source code package' by Samuel Neves.

noloader:
Thanks @c0ff.

> Adjusted and tested the patches against master: blake2-init-and-test-patches.zip

[...]

PR #416, opened 2017-05-11
Fixed non-keyed blake2x hashes when using non-standard digest sizes

c0ff:
blake2: Fixed initialization when using non-keyed constructor with non-standard digest size. Added test vectors.

https://github.com/weidai11/cryptopp/issues/415

Test vectors generated with pyblake2 which uses 'BLAKE2 reference source code package' by Samuel Neves


noloader:
Thank you very much.

c0ff:
found another instance of NULL

noloader:
Also see [Issue 415](https://github.com/weidai11/cryptopp/issues/415).

noloader:
> found another instance of NULL

That's fine. I was going to clear them for you. You can make another PR if you'd like.

PR #424, opened 2017-05-16
CMake: build RDRAND only for X86/X32/X64

### LINKED ISSUE #419: armv7l/aarch64: error: ‘HasRDRAND’ was not declared in this scope
Built against 35451f3

Build logs:

[armv7l | GCC 5.4.0 | Ubuntu Xenial](https://build.getmonero.org/builders/kovri-all-ubuntu-arm7/builds/321/steps/compile/logs/stdio)
[aarch64 | GCC 6.2.1 | Debian 8](https://build.getmonero.org/builders/kovri-all-debian-arm8/builds/306/steps/compile/logs/stdio)

Successful build logs for remaining platforms:

https://github.com/monero-project/kovri/pull/639

anonimal:
Resolves #419.

Note: I've only tested on x86_64, i686, and armv7/8 processors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,6 +294,7 @@ list(REMOVE_ITEM cryptopp_SOURCES
 		${CMAKE_CURRENT_SOURCE_DIR}/integer.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/pch.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/simple.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/rdrand.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/winpipes.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/cryptlib_bds.cpp
 		${cryptopp_SOURCES_TEST}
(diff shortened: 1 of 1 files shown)


noloader:
Thanks @anonimal ,

I think I ran the same tests several days ago on my test systems. The thing that worries me about using `CMAKE_SYSTEM_PROCESSOR` is it fails so often it not reliable. I'm pretty sure I got the *"unknown"* answer more often than I got a useful answer.

anonimal:
Ok, a few questions:

1. On the systems you've tested on that report *unknown*, if CMake reported accurately, would it match the regexp?
2. Since no one else seems to be complaining about #419, how many active cryptopp CMake users could there be?
3. If there are not many or any cryptopp CMake users (anymore), should this PR be merged and we take chances for bad reporting or should I stop using CMake with cryptopp all-together (so I can get my builds going again)?

noloader:
@anonimal,

Maybe the easiest/best way to proceed is to sidestep Cmake altogether. That is, just provide an implementation that throws on ARM or MIPS. We won't have to worry what Cmake does (or does not) do.

noloader:
@anonimal,

We went with an end-around. We avoid the problem by always building RDRAND and RDSEED, even on platforms that don't have it like ARM and MIPS. We don't have to worry about Cmake providing incorrect or incomplete information.

Also see [Commit 1b1c32d7cf88](https://github.com/weidai11/cryptopp/commit/1b1c32d7cf883078e0da02aaa38981f81c639cf3)

anonimal:
Thanks @noloader.

PR #425, opened 2017-05-18
hrtimer.cpp: fix Win32 DLL-Import build

MarcelRaad:
PerformanceCounterFrequency is used independently of CRYPTOPP_IMPORTS.
This fixes the DLL-Import Visual Studio solution configurations.

### DIFF
--- a/hrtimer.cpp
+++ b/hrtimer.cpp
@@ -30,8 +30,6 @@
 
 NAMESPACE_BEGIN(CryptoPP)
 
-#ifndef CRYPTOPP_IMPORTS
-
 #if defined(CRYPTOPP_WIN32_AVAILABLE)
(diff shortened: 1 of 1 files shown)


noloader:
@MarcelRaad,

Sorry about the late reply... I don't recall compile problems with `hrtimer.cpp`. Did you experience a problem, or is this preemptive?

MarcelRaad:
I got the following compile error in the "DLL-Import" solution configurations when `THREAD_TIMER_AVAILABLE` is not defined (Windows Store/Phone/UWP apps):

```
hrtimer.cpp(164): error C3861: 'PerformanceCounterFrequency': identifier not found
```

noloader:
Ack, thank you very much.

MarcelRaad:
Thanks for merging!

PR #426, opened 2017-05-19
have constant link order

bmwiedemann:
to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Without this patch g++ would order functions in libcryptopp.so.5.6.5
depending on random order of files in the build system's filesystem.


noloader:
Thank you very much

PR #437, opened 2017-06-18
Fix C++17 build.

Romain-Geissler-1A:
Hi,

I am having issues when compiling Cryptopp with gcc 7 and -std=gnu++17. The reason is ::byte (defined by cryptopp) and std::byte (defined by C++17) are ambiguous when having a "using namespace std" statement.

This pull request fixes the very minimal things to make the build work. Yet another pull request is required to make Cryptopp usable by someone who used "using namespace std" in it's source file when including cryptopp. That will come later.

Cheers,
Romain

### DIFF
--- a/dlltest.cpp
+++ b/dlltest.cpp
@@ -11,23 +11,22 @@
 #endif
 
 USING_NAMESPACE(CryptoPP)
-USING_NAMESPACE(std)
 
 void FIPS140_SampleApplication()
 {
 	if (!FIPS_140_2_ComplianceEnabled())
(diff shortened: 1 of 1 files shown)


noloader:
@Romain-Geissler-1A,

I believe we avoided the removal of `USING_NAMESPACE(std)` back when others were converted because we were trying to preserve the FIPS-ness of that particular file. We've pretty much violated the "sequestered file" thing everywhere else, so preserving it here does not make much sense.

Thanks for the PR.

Issue #178, opened 2016-06-07
Recent Windows 8 and Windows 10 support broke Windows 7/VS 2012


noloader:
Recently we added better support for [Windows 8, Windows 10, Windows Store, and other newer Windows platforms](http://github.com/weidai11/cryptopp/issues/143). It appears Microsoft is doing something unexpected, and its causing `cryptest.exe` to fail under Windows 7/Visual Studio 2012. Running `cryptest.exe` results in: ***`The procedure entry point GetOverlappedResultEx could not be located in the dynamic link library KERNEL32.dll`***.

![windows-entry-point](https://cloud.githubusercontent.com/assets/3538226/15842555/fd57ec8c-2c27-11e6-8acd-ad3a98bcc5ea.png)

Calling the newer functions, like `GetOverlappedResultEx`, is the result of better support for the newer Windows platforms and squashing the warnings generated by the Visual Studio toolchain. See, for example, the [`USE_WINDOWS8_API` macro in `winpipes.cpp`](http://github.com/weidai11/cryptopp/blob/master/winpipes.cpp). *No good deed goes unpunished*, as they say...

We don't set `WINVER` or `_WIN32_WINNT`. We don't have a `<winver.h>` or `<targetver.h>` in our [pre-converted Visual Studio 2010 solution and project files](http://www.cryptopp.com/wiki/Visual_Studio#Visual_Studio_2010). We expect users to set `WINVER` or `_WIN32_WINNT` appropriately when they have the need; otherwise, they get the default build environment. The folks contributing to [Cmake maintenance](http://github.com/weidai11/cryptopp/blob/master/CMakeLists.txt#L55) do this for their Windows UWP builds.

Its not clear to me why Microsoft is setting Windows 8 for a Windows 7 build and execution environment. We are trying to gather details on Stack Overflow at [Visual Studio setting WINVER/_WIN32_WINNT to Windows 8 on Windows 7?](http://stackoverflow.com/q/37668692)

Its not clear to me if we can easily clear this error. We ***can't*** do something like below because Microsoft's toolchain is selecting the incorrect value, so our code will never execute:

```
#ifdef CRYPTOPP_WIN32_AVAILABLE
# ifndef _WIN32_WINNT
#  define _WIN32_WINNT 0x0500
# endif
#endif
```

[...]

PR #439, opened 2017-06-23
Fix Windows XP support on Visual Studio 2015+

MarcelRaad:
When compiling with Visual Studio 2015+, Crypto++ uses CryptoNG by
default. CryptoNG is only available on Windows Vista and later and
Crypto++ currently ignores if the user explicitly wants to target
Windows XP. Unlike with other Windows SDK features, everything
compiles, but the application doesn't start on Windows XP because
bcrypt.dll is missing even when setting the target Windows version correctly
at compile time. That is an issue when updating Visual Studio because the
root cause is hard to find.

Making use of CryptoNG when targeting Windows 8+ instead by default,
regardless of the Visual Studio version, to fix this.
(Using Windows 8 rather than Windows 10 because "Modern" Windows Apps
need to target at least Windows 8 and it's not possible to accidently target
Windows 8+ when building applications supporting Windows XP)

### DIFF
--- a/config.h
+++ b/config.h
@@ -44,10 +44,10 @@
 // This macro will be ignored if NO_OS_DEPENDENCE is defined.
 // #define USE_MS_CNGAPI
 
-// If the user did not make a choice, then select CryptoNG if either
-// Visual Studio 2015 is available, or Windows 10 or above is available.
+// If the user did not make a choice, then select CryptoNG if
+// targeting Windows 8 or above.
 #if !defined(USE_MS_CRYPTOAPI) && !defined(USE_MS_CNGAPI)
-# if (_MSC_VER >= 1900) || ((WINVER >= 0x0A00 /*_WIN32_WINNT_WIN10*/) || (_WIN32_WINNT >= 0x0A00 /*_WIN32_WINNT_WIN10*/))
+# if !defined(_USING_V110_SDK71_) && ((WINVER >= 0x0602 /*_WIN32_WINNT_WIN8*/) || (_WIN32_WINNT >= 0x0602 /*_WIN32_WINNT_WIN8*/))
 #  define USE_MS_CNGAPI
 # else
 #  define USE_MS_CRYPTOAPI


noloader:
@MarcelRaad,

I believe we chose VS2015 because of [Issue 178](https://github.com/weidai11/cryptopp/issues/178). We needed to push the VS limit higher to ensure we avoided a similar problem on WIndows 7.

A quick question... What happens with, say, VS2008 or VS2010 and XP combination? Will we be mostly OK?

Sorry to ask. This issue has caused so much grief for us. At compile time we only have some of the necessary information, and Microsoft does not provide the proper runtime support to ensure things "just work".

MarcelRaad:
@noloader

>I believe we chose VS2015 because of Issue 178. We needed to push the VS limit higher to ensure we avoided a similar problem on WIndows 7.
>
>A quick question... What happens with, say, VS2008 or VS2010 and XP combination? Will we be mostly OK?

Yes, the Visual Studio version shouldn't make much of a difference as they all use the same Windows SDK 7 (or 2005 and 2008 even older SDKs by default, but the Windows 7 SDK is usable there too).

>This issue has caused so much grief for us. At compile time we only have some of the necessary information, and Microsoft does not provide the proper runtime support to ensure things "just work".

Yes, supporting older Windows versions is really a pain. Even when using the XP toolset, which makes no sense at all if you don't really want to support Windows XP, supporting only Windows 7 is the default. But at least you normally get compilation errors instead of runtime errors if you set the target Windows version correctly, so you immediately know where the problem is - but not with CNG :-(

noloader:
>  But at least you normally get compilation errors instead of runtime errors if you set the target Windows version correctly, so you immediately know where the problem is - but not with CNG :-(

Yeah, this seems like the better choice.

We might also see some warnings generated for deprecated APIs, but it seems like the lesser of the two evils.

MarcelRaad:
Thanks!

PR #444, opened 2017-07-17
Fix Windows compilation issues (by RaptorFactor)

MarcelRaad:
This is PR #397, but with the commit breaking C++03, the commit making clang-cl a supported compiler, and the merge commits removed. I couldn't test the Intel C++ fix, but it looks safe. I tested the other commits with Visual Studio 2017 on Windows and the default GCC 5.4.1 as well as clang 3.8 on Ubuntu Xenial.


noloader:
Thank you very much.

Issue #445, opened 2017-07-18
Visual Studio 64 bit builds produce /arch:SSE2 warnings


reganheath:
1>cl : Command line warning D9002: ignoring unknown option '/arch:SSE2'

This is triggered by the:
      <EnableEnhancedInstructionSet>StreamingSIMDExtensions2</EnableEnhancedInstructionSet>
project setting in the group..
  <ItemDefinitionGroup Condition="'$(Platform)'=='x64'" Label="X64 Configuration">

This causes the /arch:SSE2 compile option which is only supported on x86 (32 bit).

The help text for this option says "If no option is specified, the compiler will use instructions found on processors that support SSE2." so perhaps no option should be set in either configuration..

MarcelRaad:
Unfortunately it's still required for older Visual Studio versions. I think SSE2 was made the default in Visual Studio 2012.
But the property can and should be removed from the x64 block as SSE2 is always enabled for x64 and the only thing it does is generate this warning.

MarcelRaad:
@reganheath Should be fixed with https://github.com/weidai11/cryptopp/commit/3485d2180d9bdf2edadbfbb3deee2263083461b5, can you confirm?

reganheath:
Unfortunately I cannot.  We build an older version of the project for stability reasons.  I can say that the change I made to our copy matches the change shown in the diff (excluding the blank lines fixes #446).

noloader:
Also see [Command line warning D9002: ignoring unknown option '/arch:SSE2'](https://connect.microsoft.com/VisualStudio/feedback/details/1217151) on Microsoft Connect. Its an active bug, but I believe it was fixed in VS2017. I don't recall seeing it during brief testing under VS2017.

noloader:
Can this issue be closed?

reganheath:
Yep

-----Original Message-----

noloader:
Ack, thanks.

sebas77:
hmm this is still happening to me with the latest vs17 but at least I know I don't need to enable the option

NiallHornFX:
Was this fixed in 15.3.4? Still getting this.

mgradwohl:
Please document /d2archSSE42

heartbitt:
Still showing on 2022

PR #446, opened 2017-07-19
Fix Visual Studio project file issues

MarcelRaad:
- don't enable SSE2 explicitly for x64, it's always enabled and causes
  warnings (issue #445)
- delete ignored cryptest.vcxproj.user that is automatically generated
  by Visual Studio
- remove newlines in project files that Visual Studio doesn't like and
  removes on every change to project options

### DIFF
--- a/cryptdll.vcxproj
+++ b/cryptdll.vcxproj
@@ -24,17 +24,14 @@
     <PlatformToolset>v100</PlatformToolset>
     <ConfigurationType>DynamicLibrary</ConfigurationType>
   </PropertyGroup>
-
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
-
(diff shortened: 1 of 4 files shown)


noloader:
Thanks Marcel.

This is actually by design: *"delete ignored cryptest.vcxproj.user that is automatically generated
by Visual Studio"*. If you look at the file, its a convenience item and it sets the `v` argument for `cryptest.exe v`.

So the question is, do you want that to go away? I kind of like it for testing. It means I can checkout, press <kbd>F5</kbd>, and the project builds and runs the self tests.

MarcelRaad:
Ah, sorry! No, I'll remove that change then. I thought it was added accidentally as it came from a huge commit adding all Visual Studio project files, and it's in the ignore list.

MarcelRaad:
Force-pushed a new version with this change removed.

noloader:
Thank you very much.

MarcelRaad:
Thanks for merging so quickly!

PR #450, opened 2017-07-28
sockets: fix Windows 2000 build

MarcelRaad:
Commit 4630a5dab66a0e18ec8dfc0998ac223e40b3dc13 broke compilation for
Windows 2000 and earlier as getaddrinfo was introduced in Windows XP.
Fix this by including <wspiapi.h> when targeting Windows 2000 and
earlier, which falls back to an inline implementation of getaddrinfo
when necessary.
Some MinGW flavors still target Windows 2000 by default.

Ref:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms738520.aspx,
section "Support for getaddrinfo on Windows 2000 and older versions"

### DIFF
--- a/socketft.cpp
+++ b/socketft.cpp
@@ -30,8 +30,11 @@
 # include <sanitizer/msan_interface.h>
 #endif
 
-#ifdef PREFER_WINDOWS_STYLE_SOCKETS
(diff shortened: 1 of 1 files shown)


noloader:
Thank you very much.

PR #451, opened 2017-07-28
Enforce LF line endings for shell scripts

MarcelRaad:
This tells git to check out *.sh files with LF line endings. Otherwise,
Cygwin and Windows 10's Bash cannot run them
as they complain about the extra CR characters.


noloader:
Thank you very much.

Issue #466, opened 2017-08-24
Win32/64: error: inlining failed in call with (and without) -march=native


anonimal:
When building https://github.com/weidai11/cryptopp/commit/b3cacd8287e103e68f81b21a507fad2efd5f10ec *with* `-march=native` on Win32/64:

- [Win32: gcc 6.2.0](https://build.getmonero.org/builders/kovri-all-win32/builds/537/steps/compile/logs/stdio)
- [Win64: gcc 7.2.0](https://build.getmonero.org/builders/kovri-all-win64/builds/507/steps/compile/logs/stdio)

anonimal:
Same machines *without* `-march=native`:

[Win32: gcc 6.2.0](https://github.com/weidai11/cryptopp/files/1247824/Win32_cryptopp.txt)
[Win64: gcc 7.2.0](https://github.com/weidai11/cryptopp/files/1247825/Win64_cryptopp.txt)

noloader:
@anonimal,

What does `g++ -dumpmachine` return on MinGW?

anonimal:
>What does g++ -dumpmachine return on MinGW?

- `i686-w64-mingw32`
- `x86_64-w64-mingw32`

>Does this change look about right: Commit babcc1f0b705?

Building against this revision produces the same results.

noloader:
Thanks @anonimal,

So I tried to simplify things a bit at [Commit 48ec6946d729](https://github.com/noloader/cryptopp/commit/48ec6946d729cd6071bd121c738b5abf96e0e6a4).

If that fails, would you be able to add a few `message(...)` around [line 400 of CMakeList.txt](https://github.com/noloader/cryptopp/blob/master/CMakeLists.txt#L397). I'd like to ensure we are entering the two `if` blocks:

```
# New as of Pull Request 461, http://github.com/weidai11/cryptopp/pull/461.
if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"))

   message(STATUS "Clang or GNU")

   if (("${CRYPTOPP_AMD64}" STREQUAL "1") OR ("${CRYPTOPP_I386}" STREQUAL "1") OR ("${CRYPTOPP_X32}" STREQUAL "1"))

       message(STATUS "Intel IA32")
      ...
    endif()
endif()
```

noloader:
[...]

PR #471, opened 2017-08-26
CMake: create DumpMachine function for machine targeting

anonimal:
Also fixes/resolves MinGW build (Issue 466)


anonimal:
Referencing #466.

anonimal:
Note: `egrep` specifically does *not* work in this case (with MSYS2/MinGW).

noloader:
Thanks you very much

PR #478, opened 2017-08-29
Remove WORKAROUND_MS_BUG_Q258000 from the config and osrng.

wyattoday:
Remove WORKAROUND_MS_BUG_Q258000 from the config and osrng. Firstly, it's no longer used anywhere. Next, when it was used, that workaround only applied to versions of Windows prior to Windows 2000. I.e. ancient, unsecure, unused versions of Windows.

In fact, googling that "Q258000" reference number no longer returns any results because Microsoft has long since removed references to the now outdated bug.

### DIFF
--- a/config.h
+++ b/config.h
@@ -145,11 +145,6 @@
 #	error namespace support is now required
 #endif
 
-// Define this to workaround a Microsoft CryptoAPI bug where
-// each call to CryptAcquireContext causes a 100 KB memory leak.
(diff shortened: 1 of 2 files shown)


noloader:
Thanks. It was merged at [Commit 862458794363](https://github.com/weidai11/cryptopp/commit/862458794363514e9ca37c78de24b4a22b7839e1), but Git did not give your credit for it. Git absolutely sucks as a source control tool.

PR #483, opened 2017-08-30
Fix build on FreeBSD 10.3 x86 with clang++ 3.4.1.

wyattoday:
Fix build on FreeBSD 10.3 x86 with clang++ 3.4.1. The x64 build (also clang++ 3.4.1) doesn't require `CRYPTOPP_DISABLE_SHA_ASM`. It seems to be a bug specific to the x86 version of clang++.

```
clang++ -m32 -v
```

Returns the following:

```
FreeBSD clang version 3.4.1 (tags/RELEASE_34/dot1-final 208032) 20140512
Target: i386-unknown-freebsd10.3
Thread model: posix
Selected GCC installation: 
```

Also, this patch assumes this bug is fixed in clang++ (x86) 3.5. I don't actually know if that's true. But I figure it's better to version limit the disabling of the assembly, rather than always disabling the sha ASM for all x86 clang versions.

### DIFF
--- a/sha.cpp
+++ b/sha.cpp
@@ -43,7 +43,8 @@
 
 // Clang 3.3 integrated assembler crash on Linux
 //  http://github.com/weidai11/cryptopp/issues/264
-#if defined(CRYPTOPP_LLVM_CLANG_VERSION) && (CRYPTOPP_LLVM_CLANG_VERSION < 30400)
+// Clang 3.4.1 (x86) crash on FreeBSD 10.3. Clang 3.4.1 (x64) works fine.
+#if defined(CRYPTOPP_LLVM_CLANG_VERSION) && ((CRYPTOPP_LLVM_CLANG_VERSION < 30400) || CRYPTOPP_BOOL_X86 && (CRYPTOPP_LLVM_CLANG_VERSION < 30500))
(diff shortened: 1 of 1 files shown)


noloader:
Thanks @wyattoday,

How do you fell about disabling it up to 3.5 for all IA32 architectures? Maybe something like:

```
#if defined(CRYPTOPP_LLVM_CLANG_VERSION) && (CRYPTOPP_LLVM_CLANG_VERSION < 30500)
```

I prefer not to split among x86, x32 and x64. I prefer to stop disabling as soon as LLVM produces a compiler that works everywhere. I've also found Clang can be a sink-hole of time. (This is based on experience with Clang where I was adding band-aid on top of band-aid. Now I prefer to draw a line quickly even if its overly broad).

Also, for some of the BSDs the problem is in the assembler and not compiler. Some of the BSDs have an ancient GPL v1 assembler, like from Binutils 2.18.

wyattoday:
Yep, sounds good to me. I pushed your suggested change.

noloader:
Thank you very much.

noloader:
Also see [Commit ffbedcefc5d2](https://github.com/weidai11/cryptopp/commit/ffbedcefc5d2b436f5e85345ccd57e4615761b42).

PR #484, opened 2017-08-31
Make the debug build more closely behave like release if `CRYPTOPP_NO_TESTS` is defined.

wyattoday:
This commit makes it possible to make the debug build more closely behave like the release build if `CRYPTOPP_NO_TESTS` is defined.

This solves build errors if you're building the debug version of the library but don't include all of the library's functionality. For example, we use RSA but don't use PSSR. However, when using the debug version of the library (and excluding the PSSR source) we run into errors like:

```
Error	LNK2001	unresolved external symbol "private: virtual unsigned __int64 __cdecl CryptoPP::PSSR_MEM_Base::MinRepresentativeBitLength(unsigned __int64,unsigned __int64)const " (?MinRepresentativeBitLength@PSSR_MEM_Base@CryptoPP@@EEBA_K_K0@Z)
```

The other added advantage is that the debug build will now more closely match the release build (and thus reduce chance for debug/release discrepancies making bug tracking more difficult).

This change also preserves the default behavior to always build in those tests on debug builds. The user explicitly has to define `CRYPTOPP_NO_TESTS`.

### DIFF
--- a/3way.cpp
+++ b/3way.cpp
@@ -7,7 +7,7 @@
 
 NAMESPACE_BEGIN(CryptoPP)
 
-#if defined(CRYPTOPP_DEBUG) && !defined(CRYPTOPP_DOXYGEN_PROCESSING)
+#if defined(CRYPTOPP_TEST_INSTANTIATIONS)
 void ThreeWay_TestInstantiations()
 {
 	ThreeWay::Encryption x1;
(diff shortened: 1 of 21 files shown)


noloader:
Well, the library has two states: debug and release. Debug performs additional checks and tries to alert the developer of problems at development time. Release is for production software intended to be distributed.

Its not clear to me how adding an additional state is a benefit. My gut feeling is, the removal of PSSR was incomplete, and it feels like that problem should be fixed instead.

noloader:
I think we are going to pass on this request. I think it makes more sense to make it a local installation/site specific change.

PR #492, opened 2017-09-08
Check for old vs new LLVM include dir

### LINKED ISSUE #490: setenv-android.sh -> Wrong path for llvm AOSP_STL_INC
For llvm with NDK r15c, the path needs to change from
```
AOSP_STL_INC="$ANDROID_NDK_ROOT/sources/cxx-stl/llvm-libc++/libcxx/include"
```
to
```
AOSP_STL_INC="$ANDROID_NDK_ROOT/sources/cxx-stl/llvm-libc++/include"
```

I haven't opened a pull request for this since at this instant I'm not sure what version of the NDK you are meaning to support.

Deadpikle:
Fixes #490 by dynamically checking for the LLVM include directory. Defaults to the new directory. If the directory can't be found, an appropriate error is shown to the user.

I checked the beta version of the NDK -- r16 beta 1 as of this PR -- and headers appear to be in the same location there as in r15c.

### DIFF
--- a/setenv-android.sh
+++ b/setenv-android.sh
@@ -287,6 +287,12 @@ else
 	THE_STL=$(tr [A-Z] [a-z] <<< "$2")
 fi
 
+# LLVM include directory may be different depending on NDK version. Default to new location (latest NDK checked: r16beta1).
+LLVM_INCLUDE_DIR="$ANDROID_NDK_ROOT/sources/cxx-stl/llvm-libc++/include"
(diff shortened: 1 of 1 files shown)


noloader:
Thank you very much.

noloader:
@Deadpikle,,
The Android page is at https://www.cryptopp.com/wiki/Android_(Command_Line), but it does not discuss LLVM. Can you suggest some text for the LLVM runtime?

PR #531, opened 2017-11-06
RFC: workarounds for original MinGW

MarcelRaad:
- set default target Windows version to Windows XP instead of 2000 in the makefile
- if `_GLIBCXX_HAS_GTHREADS` is not defined, libstc++ doesn't provide `std::mutex` - unfortunately, the header usually used for standard library detection, `<ciso646>`, has recently been proposed to be removed from C++20 [1], so I chose `<cstddef>` instead in hopes that it would get included anyway

[1] https://wg21.link/P0619


noloader:
Thank you very much.

PR #546, opened 2017-12-15
Apply updated android build rules

### LINKED ISSUE #544: Missing compile/link flags on Android
When cross-compiling for android, one can set the API to build against by using `AOSP_API`. This sets up the system root as well include paths. However, for the latest NDK (r16) there are 2 Problems with the script:

1. You don't set the `__ANDROID_API__` for the `AOSP_FLAGS`! This needs to be set to the api version number (e.g. 21), as it is used in the android include files to enable code sections used for backward compability. Please add `-D__ANDROID_API__=<version>` to the `AOSP_FLAGS` in setenv-android.h. Otherwise it is not possible to compile against an abi older then 21.

2. A few include paths are missing. Currently, you are only setting `AOSP_STL_INC` and `AOSP_BITS_INC`, which are used as include path. But there are actually 2 more:
    - `$ANDROID_NDK_ROOT/sysroot/usr/include/<abi-specific>`. This contains additional abi specific headers. In case of arm, for example, it becomes `$ANDROID_NDK_ROOT/sysroot/usr/include/arm-linux-androideabi`
    - `$ANDROID_NDK_ROOT/sysroot/usr/include/`: A generic include path, containing many additional header definitions. Is required because it's the only place where includes like `wchar.h` can be found. The other paths only provide `cwchar` etc. Should propably be set as the last include path to search for headers.

With these 2 changes, I was able to compile with the latest ndk. I haven't tried older NDKs yet, but these changes should not break anything for older NDKs. If you need more details on what did not work without these changes, I can add more details.

Skycoder42:
fixes #544

For details, see https://android.googlesource.com/platform/ndk.git/+/HEAD/docs/UnifiedHeaders.md

Successfully tested platforms (with NDK r16):
- [ ] arm: See below
- [ ] armv7a: See below
- [ ] hard: See below
- [x] neon
- [ ] armv8: `arm_acle.h` not found.
- [x] mips
- [x] mips64
- [x] x86
- [x] x86_64

All the non-neon arm builds fail with: `#error You must enable NEON instructions (e.g. -mfloat-abi=softfp -mfpu=neon) to use arm_neon.h` This however was the case even before applying my patches, and seems to be a completly different issue on *your* current master branch. I tried to build the 5.6.5 release by patching the 2 files I changed in that release, and there they all work, so this issue is definitly unrelated to my change.

For the armv8: It seems to not exist in the "normal" NDK paths (anymore)? I checked all possible include paths using `find`, and only found those:

```
./toolchains/renderscript/prebuilt/linux-x86_64/clang-include/arm_acle.h
./toolchains/llvm/prebuilt/linux-x86_64/lib64/clang/5.0/include/arm_acle.h
./toolchains/llvm/prebuilt/linux-x86_64/lib64/clang/5.0.300080/include/arm_acle.h
./toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/include/arm_acle.h
```

Possibly yet another path that must be included?


noloader:
Thanks @Skycoder42,

> For the armv8: It seems to not exist in the "normal" NDK paths ...

This should be the easy one. I believe it was cleared at [Commit 93f690111966](https://github.com/weidai11/cryptopp/commit/93f690111966).

-----

> All the non-neon arm builds fail with: #error You must enable NEON instructions ...

This one seems a little trickier. For the ARMv7-a issue, I believe `setenv` sets `AOSP_FLAGS` properly. Or used to set them properly (in earlier NDK's). From `setenv-android.sh`:

```
# https://developer.android.com/ndk/guides/abis.html
case "$THE_ARCH" in
  arm|armv5|armv6|armv7|armeabi)
	...
	;;
  armv7a|armv7-a|armeabi-v7a)
	TOOLCHAIN_ARCH="arm-linux-androideabi"
	TOOLCHAIN_NAME="arm-linux-androideabi"
	AOSP_ABI="armeabi-v7a"
	AOSP_ARCH="arch-arm"
	AOSP_FLAGS="-march=armv7-a -mthumb -mfpu=vfpv3-d16 -mfloat-abi=softfp -Wl,--fix-cortex-a8 -funwind-tables -fexceptions -frtti"
        ;;
        ...
```

Then, `GNUmakefile-cross` uses it:

```
ifeq ($(IS_ANDROID),1)
  # CPP, CXX, AR, RANLIB, LD, etc are set in 'setenv-android.sh'
  CXXFLAGS += $(AOSP_FLAGS) -DANDROID --sysroot=$(AOSP_SYSROOT)
  CXXFLAGS += -Wa,--noexecstack -I$(AOSP_STL_INC)

  # c++config.h shows up in odd places at times.
  ifneq ($(AOSP_BITS_INC),)
    CXXFLAGS += -I$(AOSP_BITS_INC)
  endif

  # STL headers
  LDLIBS += $(AOSP_STL_LIB)

  # Source files copied into PWD for Android cpu-features
  # setenv-android.sh does the copying. Its a dirty compile.
  AOSP_CPU_OBJ = cpu-features.o
endif
```

I'm wondering if my notes at [Android | NDK CXXFLAGS](https://www.cryptopp.com/wiki/Android_(Command_Line)#armeabi-v7a) are incomplete or wrong. Our regular [GNUmakefile](https://github.com/weidai11/cryptopp/blob/master/GNUmakefile) uses ARMv7-a as a baseline for native builds (like a BananaPi or CubieTruck), and its flags are:

    FP_ABI ?= hard
    ...

    -march=armv7-a -mfloat-abi=$(FP_ABI) -mfpu=neon

I'm wondering if we should be using `-mfpu=neon` instead.

What do you think?

noloader:
Thanks @Skycoder42,

> For the armv8: It seems to not exist in the "normal" NDK paths ...

This should be the easy one. I believe it was cleared at [Commit 93f690111966](https://github.com/weidai11/cryptopp/commit/93f690111966).

-----

> All the non-neon arm builds fail with: #error You must enable NEON instructions ...

This one seems a little trickier. For the ARMv7-a issue, I believe `setenv` sets `AOSP_FLAGS` properly. Or used to set them properly (in earlier NDK's). From `setenv-android.sh`:

```
# https://developer.android.com/ndk/guides/abis.html
case "$THE_ARCH" in
  arm|armv5|armv6|armv7|armeabi)
	...
	;;
  armv7a|armv7-a|armeabi-v7a)
	TOOLCHAIN_ARCH="arm-linux-androideabi"
	TOOLCHAIN_NAME="arm-linux-androideabi"
	AOSP_ABI="armeabi-v7a"
	AOSP_ARCH="arch-arm"
	AOSP_FLAGS="-march=armv7-a -mthumb -mfpu=vfpv3-d16 -mfloat-abi=softfp -Wl,--fix-cortex-a8 -funwind-tables -fexceptions -frtti"
        ;;
        ...
```

Then, `GNUmakefile-cross` uses it:

```
ifeq ($(IS_ANDROID),1)
  # CPP, CXX, AR, RANLIB, LD, etc are set in 'setenv-android.sh'
  CXXFLAGS += $(AOSP_FLAGS) -DANDROID --sysroot=$(AOSP_SYSROOT)
  CXXFLAGS += -Wa,--noexecstack -I$(AOSP_STL_INC)

  # c++config.h shows up in odd places at times.
  ifneq ($(AOSP_BITS_INC),)
    CXXFLAGS += -I$(AOSP_BITS_INC)
  endif

  # STL headers
  LDLIBS += $(AOSP_STL_LIB)

  # Source files copied into PWD for Android cpu-features
  # setenv-android.sh does the copying. Its a dirty compile.
  AOSP_CPU_OBJ = cpu-features.o
endif
```

I'm wondering if my notes at [Android | NDK CXXFLAGS](https://www.cryptopp.com/wiki/Android_(Command_Line)#armeabi-v7a) are incomplete or wrong. Our regular [GNUmakefile](https://github.com/weidai11/cryptopp/blob/master/GNUmakefile) uses ARMv7-a as a baseline for native builds (like a BananaPi or CubieTruck), and its flags are:

    FP_ABI ?= hard
    ...

    -march=armv7-a -mfloat-abi=$(FP_ABI) -mfpu=neon

I'm wondering if we should be using `-mfpu=neon` instead.

What do you think?

Skycoder42:
I am currently trying to get travis running for android - this way we can get more reliable results.

noloader:
> I am currently trying to get travis running for android

That would be nice. ARM and IA-32 would be helpful because the Android toolchain is "different enough" to have trouble on occasion. I tired to do similar in the past but AppVeyor or Travis said they were not adding Android NDK support.

Skycoder42:
No official support, but by using the sudo environment we can install it. See [install-android.sh](https://github.com/Skycoder42/cryptopp/blob/master/TestScripts/install-android.sh).

Skycoder42:
Status update:
- Travis CI is working, results can be found here: https://travis-ci.org/Skycoder42/cryptopp
- currently, ALL llvm builds are failing with the same error. See [neon llvm](https://travis-ci.org/Skycoder42/cryptopp/jobs/317328096#L1003) as example
- armv5 and armv7 non neon builds are still failing. Maybe neon needs to be enabled for all arms in the newer ndk? Where did you get that flags from?
- armv8 now gives another error, some undeclared crc methods in `bool CryptoPP::CPU_ProbeCRC32()`

I disabled all non-android builds for now, to speed up the ci builds

noloader:
> Where did you get that flags from?

I set up a native/JNI "hello world" project. Then, I used `ndk-build V=1` to view the compiler options used by the build system for the NDK

I'm thinking we should simplify the matrix in `setenv-android.sh` a bit. Android uses soft floats, so we can probably remove the hard float stuff.

-----

> currently, all llvm builds are failing with the same error. See neon llvm as example

I would not worry too much about LLVM. The last time I checked it was still mostly experimental. The information came from one of the AOSP teams that commented on a bug report I filed about C++11 LLVM and STLport support.

-----

> armv5 and armv7 non neon builds are still failing

We should probably drop armv5 [attempted] support. It has never really been tested, and all modern devices have NEON to speed up streaming rich content. I have not seen a device without NEON (to date).

Here is what I would recommend. For the armv5 and armv6 `setenv-android.sh` code paths, print an error and exit with a code of 1. That will avoid silent failures, and signal special work needs to be done for armv5 and armv6.

armv7 is kind of surprising, especially with `-mfpu=neon`. I'll need to see an error message.

------

> armv8 now gives another error, some undeclared crc methods in bool CryptoPP::CPU_ProbeCRC32()

It would probably be a good idea to provide an error message. I think I remember something about this. CRC32 and CRC32C is part of core instruction set for the ARMv8-a profile. I believe Android is missing toolchain support for CRC, so we had to disable the code paths by undefining `CRYPTOPP_ARM_CRC32_AVAILABLE`.

Skycoder42:
I did not copy error messages, as they are available in travis. Here are direct links to the errors, where you can see the full logs:
- armv7a without neon error begins at: https://travis-ci.org/Skycoder42/cryptopp/jobs/317386371#L657
- armv7a-hard fails with the same error as armv7a, so I think the hard part is not the problem. See here: https://travis-ci.org/Skycoder42/cryptopp/jobs/317386371#L657
- armv7a-neon does work as expected. It's only the non-neon builds that fail
- armv8 errors start here: https://travis-ci.org/Skycoder42/cryptopp/jobs/317386373#L617

After exluding llvm and armv5 from the travis tests, all other platforms build successfully.

------

> I have not seen a device without NEON (to date).

Should should make all armv7 builds use neon? This would at least solve these problems.

------

> By the way, if the message is similar to error: '__crc32w' was not declared in this scope, then that's because the declaration is in <arm_acle.h> but AOSP does not supply it. Maybe we need to declare it as extern for AOSP.

Thats exactly the case. You can check the logs youself, as stated above.

noloader:
Thanks again @Skycoder42.

A quick peek at the output:

```
arm-linux-androideabi-g++ -DNDEBUG -g2 -O3 -fPIC -pipe -D__ANDROID_API__=21 -mar
ch=armv7-a -mfpu=neon -mfloat-abi=softfp -Wl,--fix-cortex-a8 -funwind-tables -fe
xceptions -frtti -DANDROID --sysroot=/home/travis/android/sdk/ndk-bundle/sysroot
 -Wa,--noexecstack -I/home/travis/android/sdk/ndk-bundle/sources/cxx-stl/gnu-lib
stdc++/4.9/include -I/home/travis/android/sdk/ndk-bundle/sysroot/usr/include/arm
-linux-androideabi -I/home/travis/android/sdk/ndk-bundle/sysroot/usr/include/ -I
/home/travis/android/sdk/ndk-bundle/sources/cxx-stl/gnu-libstdc++/4.9/libs/armea
bi-v7a/include -c cryptlib.cpp
```

`AOSP_SYSROOT` does not quite look right. I think something is off in the way Qt sets up their flags. I think it should be something like `/home/travis/android/sdk/ndk-bundle/platforms/android-21/arch-arm/`. The `include/` and `lib/` directors will be found/used at `$AOSP_SYSROOT/usr/inlcude` and `$AOSP_SYSROOT/usr/lib`. I think it is due to this:

    export AOSP_SYSROOT="$ANDROID_NDK_ROOT/sysroot"

And then later, this does not look quite right to me:

    LDFLAGS += --sysroot=$(AOSP_LD_SYSROOT)

`AOSP_SYSROOT` should handle the system's root file directory, and in this instance, the same flags should be used for compile and link. I've never experienced a need to do more than `CXXFLAGS += --sysroot=...`.

The `AOSP_LD_SYSROOT` looks unneeded. We should be using `--sysroot=$AOSP_SYSROOT` to bring in the header files and libraries.

The `-I/home/travis/android/sdk/ndk-bundle/sysroot/usr/include/` should not be needed. I think something is off in the way Qt sets up their flags. The path is part of `--sysroot`. 

We use the same compiler options (`CXXFLAGS`) for compile and link, so everything should work as expected. We don't want to start doing some things others are doing, like calling the linker directly. Best practice according to the compiler guys is to pass everything through the compiler driver, and let the compiler setup libraries.

All in all I think Android has things pretty much fucked up, from their build system through their directory structures and file system layout to their documentation. The AOSP build system has always been awful, and that's why I moved to this sort of cross compile. Recently I learned of the fucked up filesystem layout when trying to figure out how to use `cpu-features` on for a target. I did not realize we needed to compile from sources, and I did not know where to find it because artifacts show up everywhere in their filesystem layout.

There's some reading on `--sysroot` at [3.15 Options for Directory Search](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html) in the GCC manual.

Skycoder42:
> AOSP_SYSROOT does not quite look right. I think something is off in the way Qt sets up their flags.

Actually - no. Sorry I gave you the idea this has anything to do with Qt. I only used it in the beginning to figure out what is different between those two to find a lead on what could be wrong. I did *not* get these changes from Qt, but only from the official [NDK documentation](https://android.googlesource.com/platform/ndk.git/+/HEAD/docs/UnifiedHeaders.md). There it is stated:
- The compile time sysroot is now `$NDK/sysroot`. Previously this was `$NDK/platforms/android-$API/arch-$ARCH`
- At link time, change nothing. All link time build behavior should match the deprecated headers behavior. `--sysroot` should still point to `$NDK/platforms/android-$API/arch-$ARCH/`

So yeah, android is pretty weird, because we need different sysroots for compile/link. I agree with you that this is strange and you should not do this, but well, it's android... Thats why I changed the primary sysroot, and why I added the `AOSP_LD_SYSROOT`. I assume gcc behaves for the `--sysroot` as for most flags - later occurrences override previous ones, thus gcc should prefer the second sysroot for linking.

If you want, I can change it back. As long as I have the explicit `$NDK/sysroot/usr/include`, I could not obverse any difference between the two sysroots at compile time. If we keep this, yes, we can remove the explicit `$NDK/sysroot/usr/include`. Whatever you would prefer.

Skycoder42:
So, should I change this back, or can we keep it?

Once you think this PR is ready to be merge, tell me, so I can squash all the CI-commits together to keep the history clean.

noloader:
@Skycoder42,

I updated the old Android script at [Commit 5cee4a657357](https://github.com/weidai11/cryptopp/commit/5cee4a657357). The commit adds `-DCRYPTOPP_DISABLE_ASM` to non-NEON ARM.

Could you add the same to your scripts?

Also, if you could, update the PR with your latest rendition. I want to get it tested and included in Crypto++ 6.0.

Skycoder42:
Allright, I will have a look at it tomorrow. Would be awesome to get this before 6.0

noloader:
Thanks @Skycoder42.

Don't put yourself out because of it.

I was also thinking... You provided two useful links. Be sure to add them to the head notes in the script to save others the trouble,

Thanks again.

Skycoder42:
After including the latest changes, all platforms but `hard` build successfully!

The hard configuration fails for all stls with:
```
.../ld: error: cannot find -lm_hard
```
(See https://travis-ci.org/Skycoder42/cryptopp/jobs/331410619#L732)

I disabled the hard part for now, as you stated before that:

> I'm thinking we should simplify the matrix in setenv-android.sh a bit. Android uses soft floats, so we can probably remove the hard float stuff.

I also compressed these changes into 1 clean commit, to keep the cryptopp upstream clean. With that, I think this PR is ready for merging, unless you want anything changed.

(The latest android-only build can be found here: https://travis-ci.org/Skycoder42/cryptopp/builds/331421085)

noloader:
@Skycoder42, Thank you very much.

noloader:
@Skycoder42,

I like the way you handled Android in Travis. I think we should do the same for iOS.

noloader:
Ping @anonimal. He does a lot of work with Android.

PR #548, opened 2017-12-16
Add pkgconfig file libcrypt++.pc

zorun:
This is used by pkgconfig to know which build and link flags should be
used when compiling against crypto++.

One issue is that the version is hard-coded in the pkgconfig file.

Ideally, we would generate this file with the right version when building.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -840,6 +840,10 @@ install:
 	$(CP) *.h $(DESTDIR)$(INCLUDEDIR)/cryptopp
 	-$(CHMOD) 0755 $(DESTDIR)$(INCLUDEDIR)/cryptopp
 	-$(CHMOD) 0644 $(DESTDIR)$(INCLUDEDIR)/cryptopp/*.h
+	@-$(MKDIR) -p $(DESTDIR)$(LIBDIR)/pkgconfig
+	$(CP) libcrypto++.pc $(DESTDIR)$(LIBDIR)/pkgconfig
+	-$(CHMOD) 0755 $(DESTDIR)$(LIBDIR)/pkgconfig
+	-$(CHMOD) 0644 $(DESTDIR)$(LIBDIR)/pkgconfig/libcrypto++.pc
 ifneq ($(wildcard libcryptopp.a),)
 	@-$(MKDIR) -p $(DESTDIR)$(LIBDIR)
 	$(CP) libcryptopp.a $(DESTDIR)$(LIBDIR)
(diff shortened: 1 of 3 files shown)


noloader:
Thanks @zorun.

This one is kind of tricky for a few reasons. First, `pkg-config` leaves a lot to be desired. For example, the flags are mostly wrong. Here I am talking about the flags in general and not your choices. Also see [C++ project flags for *.pc.in file run by configure?](https://lists.freedesktop.org/archives/pkg-config/2017-November/001087.html) on the pkgconfig mailing list.

Second, the library name is different depending on the platform. Red Hat, Fedora and compatibles use `libcrypto++`. Debian, Ubuntu and compatibles use `libcryptopp`. I've never seen that before and I am not sure how to proceed. I tried to figure out a way to change it dynamically but was not able to find a solution. Also see [Abstracting include paths and library names with pkg-config?](https://stackoverflow.com/q/32920396/608639) on Stack Overflow.

Third, path names are a problem. Debian, Ubuntu and friends use `/lib` and `/usr/lib`. Red Hat, Fedora and friends use `/lib64` and `/usr/lib64` on amd64. Worse, Red Hat and Fedora `config.site` is not properly configured for alternate location like `/usr/local` or `/opt/local`.  Also see [ Document that config.site is intended for (and only for) /usr prefix, provide doc example for /usr/local/share/config.site](https://bugzilla.redhat.com/show_bug.cgi?id=1510073) in the Red Hat bug reporter.

Fourth, we have a `pkg-config` at [NoLoader | cryptopp-autotools](https://github.com/noloader/cryptopp-autotools). It is Debian-biased, meaning it uses Debian naming conventions. That's because László Böszörményi is a Debian maintainer and our packager maintainer. The Fedora EPEL maintainers use László's work as a starting point, and Fedora renames the library in their build scripts.

At this point I am not sure what to do. What do you recommend?

zorun:
I don't have any opinion on pkg-config, I just thought it would be worth it to upstream our pkgconfig file: https://git.archlinux.org/svntogit/community.git/tree/trunk/libcrypto++.pc?h=packages/crypto%2b%2b
I'm perfectly happy to keep shipping this as part of the Arch package.

noloader:
@zorun,

I'm going to close this without merging for the time being. I think Arch is in a good place since you guys have a working Autotools setup. Other folks can use Debian's work, Aarch's work or even my GitHub if they need it.

Are you OK with the close?

zorun:
@noloader yes no problem

noloader:
@zorun,

We made two small changes to the makefile. First, we added a `make libcryptopp.pc` recipe that writes a basic `*.pc` file. It is an optional step and users need to run it. We based it off of `*.pc` files like yours.

Second, if `libcryptopp.pc` is present during `make install`, then we install it along with the other files. This should avoid custom patches to makefiles for folks like you who supply the `*.pc` file already.

Also see [Issue 582, Use install program in Makefiles](https://github.com/weidai11/cryptopp/issues/582).

Issue #520, opened 2017-10-12
Borland and "E2268 Call to undefined function log"


noloader:
From [Problem building Crypto++ with C++Builder XE5 - EMBARCADERO - RAD Studio 12.0](https://groups.google.com/forum/#!topic/cryptopp-users/QGJNGD-2Pgs):

> ```
> nbtheory.cpp(1026,90): error E2268: E2268 Call to undefined function 'log'
> nbtheory.cpp(1033,90): error E2268: E2268 Call to undefined function 'log' 
> 
> [bcc32 Fatal Error] adhoc.cpp(1): F1009 Unable to open input file 'adhoc.cpp'
>  
> [TLIB Warning] Warning: library was too large for page size, rebuilt with page size 1024
> ```
> 
> But I got a lot of errors when I build the .exe
> 
> ```
> [bcc32 Error] bench1.cpp(64): E2268 Call to undefined function 'log'
> [bcc32 Error] bench1.cpp(100): E2268 Call to undefined function 'log'
> [bcc32 Warning] bench1.cpp(407): W8012 Comparing signed and unsigned values
> [bcc32 Error] factory.h(104): E2040 Declaration terminated incorrectly
> [bcc32 Error] factory.h(105): E2451 Undefined symbol 's_registry'
> [bcc32 Error] factory.h(105): E2357 Reference initialized with 'undefined', needs lvalue of type 'ObjectFactoryRegistry<RandomNumberGenerator,0>'
> [bcc32 Error] factory.h(104): E2040 Declaration terminated incorrectly
> [bcc32 Error] factory.h(105): E2451 Undefined symbol 's_registry'
> [bcc32 Error] factory.h(105): E2357 Reference initialized with 'undefined', needs lvalue of type 'ObjectFactoryRegistry<NIST_DRBG,0>'
> [bcc32 Error] factory.h(104): E2040 Declaration terminated incorrectly
> [bcc32 Error] factory.h(105): E2451 Undefined symbol 's_registry'
> [bcc32 Error] factory.h(105): E2357 Reference initialized with 'undefined', needs lvalue of type 'ObjectFactoryRegistry<HashTransformation,0>'
> [bcc32 Error] factory.h(104): E2040 Declaration terminated incorrectly
> [bcc32 Error] factory.h(105): E2451 Undefined symbol 's_registry'
[...]

PR #552, opened 2017-12-30
Change ::log() to log() to work with C++Builder

tanzislam:
As reported in #520, C++Builder standard libraries don't have a `log()` function at global namespace. Change the invocations to unqualified name lookup, and apply a using-declaration to `std::log()` when compiling under C++Builder.


noloader:
Thank you very much.

PR #560, opened 2018-01-10
Deterministic k (RFC6979): issue in bits2octets and new test case

songlinm:
I have been experimenting with ECDSA_RFC6979 by feeding data from the original RFC-6979 ([section A.1.2](https://tools.ietf.org/html/rfc6979#appendix-A.1.2)). The result calculated by DL_Algorithm_ECDSA_RFC6979::GenerateRandom does not agree with the one given by the RFC.

I have created a unit test case (see 'ValidateECDSA_RFC6979' in diff), using the same data given by the RFC. It currently fails with the latest code.

It seems to me the problem is in bits2octets. Specifically, when it calls bits2int, according to RFC [section 2.3.2](https://tools.ietf.org/html/rfc6979#section-2.3.2), the bit count should be that of the base point (i.e. "q", in order to "shrink" the hash value). The current code uses the hash value's bit count. 

By changing the bit count to q.BitCount(), the new unit test case passes - the generated "k" matches the RFC's value.

### DIFF
--- a/gfpcrypt.h
+++ b/gfpcrypt.h
@@ -384,7 +384,7 @@ protected:
     // curve's order.
     SecByteBlock bits2octets(const SecByteBlock& in, const Integer& q) const
     {
-        Integer b2 = bits2int(in, in.size()*8);
+        Integer b2 = bits2int(in, q.BitCount());
         Integer b1 = b2 - q;
         return int2octets(b1.IsNegative() ? b2 : b1, q.ByteCount());
     }
--- a/validat1.cpp
+++ b/validat1.cpp
@@ -201,6 +201,7 @@ bool ValidateAll(bool thorough)
 	pass=ValidateECP() && pass;
 	pass=ValidateEC2N() && pass;
 	pass=ValidateECDSA() && pass;
+	pass=ValidateECDSA_RFC6979() && pass;
 	pass=ValidateECGDSA(thorough) && pass;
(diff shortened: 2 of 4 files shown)


noloader:
@songlinm,

Thank you very much. I thought we had some test cases for RFC6979. That we are missing them is my fault.

Sorry again about that.

noloader:
@songlinm,

A quick FYI... I'm probably going to change the name `GenerateRandom` soon. `GenerateRandom` is used throughout the library for generating keys. I should have named it `GenerateCoefficient`, `CalculateK` or similar.

songlinm:
@noloader Many thanks. Yes, I agree GenerateCoefficient or CalculateK would be more intuitive. 

While we are on it, may I ask if we can allow ECDSA_RFC6979<>::Signer to not require a NullRNG? If I understand correctly, currently an RNG (albeit fake) is still needed for deterministic signatures (took me quite a while to figure out). This looks a bit counterintuitive. Not sure if the removal is feasible though, due to the inheritance. Thanks.

noloader:
> ... I ask if we can allow ECDSA_RFC6979<>::Signer to not require a NullRNG ...

DSA and ECDSA use `DL_ElgamalLikeSignatureAlgorithm` as a base class. We retained the same interface so `DSA_RFC6979` and `ECDSA_RFC6979` could be drop-in replacements for DSA or ECDSA signers.

What do you propose?

Issue #572, opened 2018-01-24
Upgrade RandomPool IncorporateEntropy from SHA-256 to SHA-384


P2Poker:
Reason: more entropy
Difficulty: pretty easy
Risk: minimal

**TODO:**

- [x] determine best way to split a 384-bit hash into a 256-bit key + 128-bit seed
- [x] determine whether any other instances of SHA256 could benefit from a similar upgrade
- [x] determine whether any other code is affected negatively
- [x] confirm overall impact
- [x] test
- [x] update documentation

**Theoretical example of possible modifications:**

```
void RandomPool::IncorporateEntropy(const byte *input, size_t length)
{
	//SHA256 hash; // replace with SHA384
	SHA384 hash;

	hash.Update(m_key, 32);
	hash.Update(m_seed, 16); // add the current existing seed to the hash update cycle

	hash.Update(input, length);

	// Finalize using a 384-bit (48 byte) object, and split hash across the key and seed
	//hash.Final(m_key);
	FixedSizeAlignedSecBlock<byte, 48> hashResult;
	hash.Final(hashResult);
	::memcpy(m_key,hashResult,32);
	::memcpy(m_seed,(byte*)hashResult+32,16);

	m_keySet = false;
}

/* ... */

template <class BLOCK_CIPHER>
void AutoSeededX917RNG<BLOCK_CIPHER>::Reseed(bool blocking, const byte *input, size_t length)
{
	/* BLOCK_CIPHER::BLOCKSIZE + BLOCK_CIPHER::DEFAULT_KEYLENGTH
		will be 48 bytes (384 bits) when AES-256 is used as the block cipher. */

	SecByteBlock seed(BLOCK_CIPHER::BLOCKSIZE + BLOCK_CIPHER::DEFAULT_KEYLENGTH);
	const byte *key;
	do
	{
		OS_GenerateRandomBlock(blocking, seed, seed.size());
		if (length > 0)
		{
			//SHA256 hash; // replace with SHA384
			SHA384 hash;
			hash.Update(seed, seed.size());
			hash.Update(input, length);
			hash.TruncatedFinal(seed, UnsignedMin(hash.DigestSize(), seed.size()));
		}
		key = seed + BLOCK_CIPHER::BLOCKSIZE;
	}	// check that seed and key don't have same value
[...]

PR #574, opened 2018-01-24
RNG tuneup

P2Poker:
Covers issue #572 , as well as some other updates.

Built and tested under debug and release modes, x64 and x86 platforms - all tests passed.

**Changes:**

AutoSeededRandomPool -
- Increased default seedSize from 32 bytes to 48 bytes.
- Added static constexpr member DEFAULT_SEEDSIZE = 48 (384 bits for AES-256).
- Added one additional overload of both the constructor and the Reseed member function (eliminating unnecessary DMA).
- Added constructors to match the standard library RNG interface.

AutoSeededX917RNG -
- Upgraded Reseed member function from SHA-256 to SHA-384. This generator was already performing similar behavior - namely, seeding all 384 bits when AES-256 is specified as the block cipher. After the change, the generator should be able to seed the entire cipher with valid hash bits. Should not affect 3DES etc. since the hash is truncated as needed.
- Added constructors to match the standard library RNG interface.

RandomPool -
- Optimized block processing loop.
- Upgraded IncorporateEntropy member function from SHA-256 to SHA-384.
- Combined m_seed and m_key into a single member variable (m_seed) of type FixedSizeAlignedSecBlock<byte, 48> (makes IncorporateEntropy faster/simpler).
- Replaced member variable m_pCipher with a new member variable (m_cipher) of type AES::Encryption (elimating unnecessary DMA).
- Set member variable m_keySet = true after calling m_cipher.SetKey(...) - this looks like a minor oversight.
- Added constructors to match the standard library RNG interface.

RandomNumberGenerator -
- Added class members to match the standard library RNG interface (result_type, min, max, seed/constructors, operator(), discard).

General -
- Updated documentation.
- Updated .gitignore to avoid additional Visual Studio build files, logs, and local symbol databases.


P2Poker:
Closing.
Will reopen with commits merged properly.

Issue #576, opened 2018-01-27
Instruction set compilation flags should always be enabled for SIMD implementations


zorun:
I am building crypto++ 6.0.0 for Arch Linux, and noticed that the new SIMD files are *not* built with the needed flags (`-msse4.1`, `-maes`, etc) by default.

These flags should be enabled inconditionnally:

- these flags are necessary (otherwise, build errors occur, like mentioned in the release notes or #573 );
- the needed CPU features are already checked at run-time (if I read `cpu.cpp` correctly) so there's no harm enabling all features at build-time;
- when packaging for a distro, the build-time and run-time CPU have different capabilities, so it makes no sense to test CPU features at build time.

Note that I am using the `GNUMakefile-cross` build system, because the normal `GNUMakefile` is not suitable for packaging (it sets various unwanted build flags with no simple way to disable them).

What I propose is something like this:

```
ifneq ($(IS_i686)$(IS_x86_64),00)
  ARIA_FLAG = -mssse3
  BLAKE2_FLAG = -msse4.2
  AES_FLAG = -msse4.1 -maes
  # etc
endif
```
that is, remove all feature detection like `$(shell echo | $(CXX) -x c++ $(CXXFLAGS) -mssse3 -mpclmul -dM -E - 2>/dev/null | $(EGREP) -i -c __PCLMUL__ )`

I tested doing just that and building on an old Core2 (which certainly doesn't have AES, SHA and so on).  It builds fine and all tests pass, printing the following:

```
hasSSE2 == 1, hasSSSE3 == 1, hasSSE4.1 == 1, hasSSE4.2 == 0, hasAESNI == 0, hasCLMUL == 0, hasRDRAND == 0, hasRDSEED == 0, hasSHA == 0, isP4 == 0
```
so runtime feature detection seems to work fine.

noloader:
Thanks @zorun.

> I am building crypto++ 6.0.0 for Arch Linux, and noticed that the new SIMD files are not built with the needed flags (-msse4.1, -maes, etc) by default.

Hmm... I'm seeing the flags added to the SIMD files on my Arch Linux version 233 test machine.

-----

> Note that I am using the GNUMakefile-cross build system, because the normal GNUMakefile is not suitable for packaging

OK, you should be using `GNUmakefile`. `GNUmakefile` works as expected on my Arch64 machine.

`GNUmakefile-cross` is for Android, some ARM and iOS.

-----

> What I propose is something like this ...

[...]

PR #577, opened 2018-01-27
build: add missing flags for SIMON and SPECK in GNUMakefile-cross

zorun:
This fixes build on Arch Linux.  The relevant bits were simply copied from GNUMakefile.


noloader:
Ugh, thanks. Simon and Speck are missing.

noloader:
Also see [Issue 576](https://github.com/weidai11/cryptopp/issues/576).

PR #589, opened 2018-02-20
Freeze ABI compatibility with LIB_MAJOR

kwizart:
This is a convention that binary compatibity uses one number.
Using that, it's possible to have bugfixes releases (patchlevel
incremented) and enhancement release (minor incremented with no
public interface removed).

Here is more information about convention
https://autotools.io/libtool/version.html
(libtool isn't relevant to this project, but the explanation hold)

Signed-off-by: Nicolas Chauvet <kwizart@gmail.com>

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -688,8 +688,8 @@ endif
 ifeq ($(HAS_SOLIB_VERSION),1)
 # Full version suffix for shared library
 SOLIB_VERSION_SUFFIX=.$(LIB_MAJOR).$(LIB_MINOR).$(LIB_PATCH)
-# Different patchlevels are compatible, minor versions are not
-SOLIB_COMPAT_SUFFIX=.$(LIB_MAJOR).$(LIB_MINOR)
+# Different patchlevels and minors are compatible since 6.1
+SOLIB_COMPAT_SUFFIX=.$(LIB_MAJOR)
 SOLIB_FLAGS=-Wl,-soname,libcryptopp.so$(SOLIB_COMPAT_SUFFIX)
 endif # HAS_SOLIB_VERSION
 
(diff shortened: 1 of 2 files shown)


noloader:
Thanks @kwizart.

You might also be interested in [Crypto++ | Autotools](https://github.com/noloader/cryptopp-autotools).

kwizart:
Thx for merging.
The problem is that this an out of tree build system. Whereas I dramatically prefers autotools over single Makefiles, I think it's easier to only maintain one build system over many.
Anything preventing from merging the autotools work within cryptopp directly instead ?

noloader:
@kwizart, @gcsideal,

Something else we do in the GNUmakefile (not the cross makefile) is create the links during install. However, according to Stallman in [GNU Coding Standards | 7.2.2 Utilities in Makefiles](https://www.gnu.org/prep/standards/standards.html#Utilities-in-Makefiles):

> It is a good idea to avoid creating symbolic links in makefiles, since a few file systems don’t support them. 

As packagers and maintainers, what do you guys recommend? Should the Makefile and our Autotools project stop creating them?

noloader:
@kwizart,

> Anything preventing from merging the autotools work within cryptopp directly instead ?

It is still on the experimental side of things, and it still has some work to be done. Check out the head notes at the top of the `configure.ac` and `Makefile.am` files. I also used a lot of shell scripting where m4 macros are supposed to be used.

I think guys like you and László  will manage, but other users will have problems. Two things I learned from the [Crypto++ | Cmake](https://github.com/noloader/cryptopp-cmake) experiment was:

1. don't expect users to work through configure problems.
2. the bug reporter will fill up if something is only working half-ass

PR #594, opened 2018-03-04
Define AT_HWCAP/AT_HWCAP2 if getauxval unavailable

ffontaine:
If CRYPTOPP_GETAUXV_AVAILABLE is undefined, getauxval function is
defined to return 0 however AT_HWCAP and AT_HWCAP2 are not defined so
compilation on toolchain without getauxval and these variables such as
uclibc-ng will fail.

### DIFF
--- a/cpu.cpp
+++ b/cpu.cpp
@@ -30,6 +30,12 @@
 #if CRYPTOPP_GETAUXV_AVAILABLE
 # include <sys/auxv.h>
 #else
+#ifndef AT_HWCAP
+#define AT_HWCAP 16
+#endif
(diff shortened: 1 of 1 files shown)


noloader:
Thanks @ffontaine.

Sorry about the failure. How can I setup a test case for this?

ffontaine:
I found this one when trying to enable the "target" version of cryptopp in buildroot. Compilation failed for ARM with uclibc-ng toolchain. buildroot can be used to build custom embedded images with different toolchains (glibc, musl, uclibc-ng) fora wide variety of targets. As soon as my patch is ready, I will be happy to provide you a link.

PR #598, opened 2018-03-10
Fix GCC version for RDSEED intrinsic

ffontaine:
rdseed is only available on gcc 4.8.x, not on gcc 4.7.x

Fixes:
 - http://autobuild.buildroot.org/results/9ab386124e4a09b50598c6f95ad40b25a83d227e


noloader:
Thank @ffontaine.

My apologies for that. I'm not sure why testing did not catch it. We test under GCC 4.7.

Issue #596, opened 2018-03-07
Calling multiple ZlibDecompressor in parallel causes adler32 checksum failure


rocksonhead:
msvc 15, 17 and gcc 5.4 64bit compilers
crypto versions 5.6.5 and 6.1
built via cmake on gcc 5.6.5
and vs projects on 5.6.5 and 6.1 - though changing runtime library to /MD
Please see attached project for example code to induce failure on windows.

The failure may not occur on the first run, but running the application several times an exception is thrown in the decompression "Adler32 check error"

The example code uses openmp to parallelize, however it can also be replicated using tbb

[test.zip](https://github.com/weidai11/cryptopp/files/1788987/test.zip)

rocksonhead:
The failure seems to be occurring somewhere around the HuffmanDecoder::Decode
This function seems to be returning 0 lengths.
The message is not printed when performing 1 decompression or in debug mode.

`inline unsigned int HuffmanDecoder::Decode(code_t code, /* out */ value_t &value) const
{
	CRYPTOPP_ASSERT(m_codeToValue.size() > 0);
	LookupEntry &entry = m_cache[code & m_cacheMask];
	code_t normalizedCode = 0;
    if (entry.type != 1)
    {
        normalizedCode = BitReverse(code);
    }

    if (entry.type == 0)
    {
        FillCacheEntry(entry, normalizedCode);
    }

	if (entry.type == 1)
	{
		value = entry.value;
        if (entry.len == 0)
            std::cout << "\nDecode failed" << std::flush;
		return entry.len;
	}
	else
	{
		const CodeInfo &codeInfo = (entry.type == 2)
			? entry.begin[(normalizedCode << m_cacheBits) >> (MAX_CODE_BITS - (entry.len - m_cacheBits))]
			: *(std::upper_bound(entry.begin, entry.end, normalizedCode, CodeLessThan())-1);
		value = codeInfo.value;
		return codeInfo.len;
	}
}`

rocksonhead:
Appreciate it!
Will continue to try and diagnose whats happening in the meantime.
Kind regards

rocksonhead:
It might be associated with the struct LookupEntry.  So far I cannot replicate the problem with an extra param in the struct.....

`	struct LookupEntry
	{
		unsigned int type;
        unsigned int hits = 0;
		union
		{
			value_t value;
			const CodeInfo *begin;
		};
		union
		{
			unsigned int len;
			const CodeInfo *end;
		};
	};`

[...]

PR #600, opened 2018-03-20
Fix calling multiple ZlibDecompressor in parallel causes adler32 chec…

rocksonhead:
…ksum failure (GH #596)

### DIFF
--- a/zinflate.cpp
+++ b/zinflate.cpp
@@ -612,41 +612,51 @@ void Inflator::FlushOutput()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


noloader:
Ping @denisbider, @mouse07410 and @denisbider.

Any objections?

denisbider:
@noloader: Looks reasonable to me.

mouse07410:
No objections. The logic of the proposed fix is convincing.

noloader:
Thank you very much.

Issue #610, opened 2018-03-28
Add KeyDerivationFunction interface


noloader:
This has been a long time coming... We have not really moved forward into the newer password hashing algorithms because we lacked the infrastructure to do so. In the old days we only needed to manage salts and iteration counts. Nowadays we have several additional parameters, like hardness factors.

The additional parameters beg for a `KeyDerivationFunction` interface that accepts a `NameValuepairs` to pass arbitrary parameters in an extensible way. While not readily apparent, a `KeyDerivationFunction` interface also provides unified testing of the algorithms using test vectors. The base interface allows us to provide factory methods to create a KDF without the need of knowing a particular constructor. The KDF is then configured using `NameValuepairs`.

This ticket will track the addition of the `KeyDerivationFunction` interface and the associated changes for existing KDF's and PBKDF's.

noloader:
Here's the first cut on the interface based on applying it to HKDF class.

```
/// \brief Interface for key derivation functions
/// \since Crypto++ 6.2
class CRYPTOPP_DLL CRYPTOPP_NO_VTABLE KeyDerivationFunction : public Algorithm
{
public:
	virtual ~KeyDerivationFunction() {}

	virtual const Algorithm & GetAlgorithm() const =0;

	/// \brief Provides the name of this algorithm
	/// \return the standard algorithm name
	virtual std::string AlgorithmName() const =0;

	/// \brief Determine minimum number of bytes
	/// \returns Minimum number of bytes which can be derived
	virtual unsigned int MinDerivedLength() const;

	/// \brief Determine maximum number of bytes
	/// \returns Maximum number of bytes which can be derived
	virtual unsigned int MaxDerivedLength() const;

	/// \brief Returns a valid key length for the derivation function
	/// \param keylength the size of the derived key, in bytes
	/// \returns the valid key length, in bytes
	virtual size_t GetValidDerivedLength(size_t keylength) const =0;

	/// \brief Returns whether keylength is a valid key length
	/// \param keylength the requested keylength
	/// \return true if the derived keylength is valid, false otherwise
	/// \details Internally the function calls GetValidKeyLength()
	virtual bool IsValidDerivedLength(size_t keylength) const {
		return keylength == GetValidDerivedLength(keylength);
	}

	/// \brief Derive a key from secret
	/// \param derived the output byte buffer
	/// \param derivedLen the size of the output byte buffers, in bytes
	/// \param seed the input byte buffer
[...]

PR #611, opened 2018-03-30
Add KeyDerivationFunction interface (GH #610)

noloader:
Also see [Add KeyDerivationFunction interface](https://github.com/weidai11/cryptopp/issues/610).


noloader:
Also see [Issue 610](https://github.com/weidai11/cryptopp/issues/610).

Issue #636, opened 2018-04-10
OS X compile failures with Clang from Xcode 9.3 for 32-bit


alanbirtles:
`c++ -c sha.cpp -arch i386` is failing to build with the following error:
```
sha.cpp:1005:5: error: invalid operand for instruction
    ASJ(    jnz,     0, b)
    ^
./cpu.h:623:23: note: expanded from macro 'ASJ'
        #define ASJ(x, y, z) GNU_ASJ(x, y, z)
                             ^
./cpu.h:617:27: note: expanded from macro 'GNU_ASJ'
        #define GNU_ASJ(x, y, z) #x " " #y #z ";" NEW_LINE
                                 ^
<scratch space>:105:2: note: expanded from here
"jnz"
 ^
<inline asm>:80:1: note: instantiated into assembly here
jnz 0b;
^
```
64-bit builds work. The clang version is:
`Apple LLVM version 9.1.0 (clang-902.0.39.1)`
vmac.cpp fails with the same error.
I presume this is a bug with clang?

noloader:
Thanks @alanbirtles,

@mouse07410 is experiencing a similar problem. I don't have access Xcode 9.3 so I have not been able to duplicate it.

I suspect it has something to do with the integrated assembler and whitespace. Also see [cpu.h : 40](https://github.com/weidai11/cryptopp/blob/master/cpu.h#L42):

```
// Applies to both X86/X32/X64 and ARM32/ARM64
#if defined(CRYPTOPP_LLVM_CLANG_VERSION) || defined(CRYPTOPP_APPLE_CLANG_VERSION) || defined(CRYPTOPP_CLANG_INTEGRATED_ASSEMBLER)
	#define NEW_LINE "\n"
	#define INTEL_PREFIX ".intel_syntax;"
	#define INTEL_NOPREFIX ".intel_syntax;"
	#define ATT_PREFIX ".att_syntax;"
	#define ATT_NOPREFIX ".att_syntax;"
#elif defined(__GNUC__)
	#define NEW_LINE
	#define INTEL_PREFIX ".intel_syntax prefix;"
	#define INTEL_NOPREFIX ".intel_syntax noprefix;"
	#define ATT_PREFIX ".att_syntax prefix;"
	#define ATT_NOPREFIX ".att_syntax noprefix;"
#else
	#define NEW_LINE
	#define INTEL_PREFIX
	#define INTEL_NOPREFIX
	#define ATT_PREFIX
[...]

PR #640, opened 2018-04-12
Workaround for #636

alanbirtles:
Temporarily switch to AT&T syntax in ASJ for clang 5.0.0+ and apple clang 9.0.0+ to workaround https://bugs.llvm.org/show_bug.cgi?id=36144

### DIFF
--- a/cpu.h
+++ b/cpu.h
@@ -614,7 +614,12 @@ inline int GetCacheLineSize()
 	#define GNU_AS2(x, y) #x ", " #y ";" NEW_LINE
 	#define GNU_AS3(x, y, z) #x ", " #y ", " #z ";" NEW_LINE
 	#define GNU_ASL(x) "\n" #x ":" NEW_LINE
+// clang 5.0.0 and apple clang 9.0.0 don't support numerical backward jumps
+#if (CRYPTOPP_LLVM_CLANG_VERSION >= 50000) || (CRYPTOPP_APPLE_CLANG_VERSION >= 90000)
+	#define GNU_ASJ(x, y, z) ATT_PREFIX ";" NEW_LINE #x " " #y #z ";" NEW_LINE INTEL_PREFIX ";" NEW_LINE
+#else
 	#define GNU_ASJ(x, y, z) #x " " #y #z ";" NEW_LINE
+#endif
 	#define AS1(x) GNU_AS1(x)
 	#define AS2(x, y) GNU_AS2(x, y)
 	#define AS3(x, y, z) GNU_AS3(x, y, z)


mouse07410:
> Clang has lots of problems with Intel syntax

:-) Yeah, tell me. ;-)

> We might want to investigate moving Clang off of intel syntax altogether. Otherwise, I think we will be patching for Clang for years to come.

I don't mind moving off of the Intel syntax.

Regarding this workaround: it works OK with (at least doesn't break) clang, but still fails with Macports GCC:
```
. . . . .
g++ -maes -mpclmul -mrdrnd -msse2 -mssse3 -msse4.2 -mtune=native -Os -Ofast -std=c++11 -fPIC -Wa,-q -DCRYPTOPP_CLANG_INTEGRATED_ASSEMBLER=1 -pthread -pipe -c rdrand.cpp
rdrand.cpp:192:1: error: invalid operand for instruction
jnc 1b;
^
rdrand.cpp:192:1: error: invalid operand for instruction
jnc 1b;
^
rdrand.cpp:349:1: error: invalid operand for instruction
jnc 1b;
^
rdrand.cpp:349:1: error: invalid operand for instruction
jnc 1b;
^
rdrand.cpp:192:1: error: invalid operand for instruction
jnc 1b;
^
rdrand.cpp:192:1: error: invalid operand for instruction
jnc 1b;
^
rdrand.cpp:349:1: error: invalid operand for instruction
jnc 1b;
^
rdrand.cpp:349:1: error: invalid operand for instruction
jnc 1b;
^
make: *** [rdrand.o] Error 1
```

Do you want this PR merged?

mouse07410:
Also, from my inline assembly experiments:
```
$ cat t-asm.c
. . . . .
  uint16_t buffer[8], *ptr = buffer;
  memset(buffer, 0x0, sizeof(buffer));
  for (size_t i = 0; i < sizeof(buffer)/2; i++, ptr++) {
    __asm__ __volatile__("rdrand %0; setc %1"
			 :"=r"(*ptr), "=qm"(err)
			 :
			 );
    printf("%04x", buffer[i]);
  }
. . . . .
$ gcc -Wa,-q -o t-asm t-asm.c
$ ./t-asm
Random buffer (16 bytes): 3e0a4aaab8dcbf9dfa0adc85fdefd830
$ 
```

mouse07410:
I suspect this workaround would work for Apple GCC, but as I demonstrated - not for Macports GCC (which is more advanced).

mouse07410:
Also, I found (following the references like https://trac.macports.org/ticket/51642#comment:10) that setting env variable `AS_INTEGRATED_ASSEMBLER=1` has the same effect on Macports GCC as `-Wa,-q` flag.

P.S. There's a good chance this env var will help me with Haskell compiler as well, but that's a different story. ;-)

mouse07410:
Just tested it again with Xcode-9.3.1 (and Macports-installed GCC-7.3.0). it works, so I'm merging it.

Note that my `CXXFLAGS` now include `-DAS_INTEGRATED_ASSEMBLER=1` in addition to `-Wa,-q`.

noloader:
> Note that my CXXFLAGS now include -DAS_INTEGRATED_ASSEMBLER=1 in addition to -Wa,-q.

Thanks @mouse07410,

Does the `GNUmakefile` need a `CXXFLAGS += AS_INTEGRATED_ASSEMBLER=1`?

mouse07410:
`CRYPTOPP_CLANG_INTEGRATED_ASSEMBLER` is what we defined ourselves. `AS_INTEGRATED_ASSEMBLER` is an **env variable** what Xcode (and Macports compilers?) seems to use.

I misspoke in my comment above - it's not a C define. But as I said, now my `~/.bashrc` includes 
```
export AS_INTEGRATED_ASSEMBLER=1
```

PR #646, opened 2018-04-26
Fix ios build for cases where Xcode is installed in a path with spaces

PetteriHuusko:
IOS build fails if Xcode is installed in a path with spaces. Fix quoting.


noloader:
Thank you very much.

Issue #652, opened 2018-05-03
Keccak256 and SHA3 crash when input is 0 length


akovachev:
### Keccak256 and SHA3 not working as intended

* State the operating system and version (Ubutnu 17 x86_64, Windows 7 Professional x64, etc)
Mac OS

* State the version of the Crypto++ library (Crypto++ 5.6.5, Master, etc)
Crypto++ 7.0.0
* State how you built the library (Makefile, Cmake, distro, etc)
Bazel

* Show a typical command line (the output of the compiler for cryptlib.cpp)
* Show the link command (the output of the linker for libcryptopp.so or cryptest.exe)
* Show the exact error message you are receiving (copy and paste it); or

```
Assertion failed: external/cryptopp/keccak.cpp(255): Update
Trace/BPT trap: 5
```

```
Assertion failed: external/cryptopp/sha3.cpp(254): Update
Trace/BPT trap: 5
```

* Clearly state the undesired behavior (and state the expected behavior)

Both SHA3 and Keccak256 have valid test vectors with 0-length input as follows:

SHA3("") = A7FFC6F8BF1ED76651C14756A061D662F580FF4DE43B49FA82D80A4B80F8434A
Keccak256("") = C5D2460186F7233C927E7DB2DCC703C0E500B653CA82273B7BFAD8045D85A470

Please note that when compiled in release test vectors succeed, however in debug the assert added due to some vulnerability is changing the behavior of these two functions.

Looking at the history I see that the asserts were added due to some vulnerability listed here:
https://github.com/weidai11/cryptopp/commit/399a1546de71f41598c15edada28e7f0d616f541

Commit was made by @noloader

noloader:
Thanks @akovachev.

I'm not sure what to do with this report. I think things are working as expected.

Debug builds are intended to be run under a debugger. If there's no debugger attached then the debug version of the program should install its own trap handler like shown in [`test.cpp : 130`](https://github.com/weidai11/cryptopp/blob/master/test.cpp#L130):

```
// See misc.h and trap.h for comments and usage
#if defined(CRYPTOPP_DEBUG) && defined(UNIX_SIGNALS_AVAILABLE)
static const SignalHandler<SIGTRAP, false> s_dummyHandler;
#endif
```

The 0-length string is one of those corner cases that you deal with by pressing `c` when debugging with GDB. Otherwise, we lose the alerts associated with the assert. It effectively throws the baby out with the bath water.

[...]

PR #654, opened 2018-05-03
Add CRYPTOPP_ASSERT to iterhash so it matches Keccak and SHA3

akovachev:
As described in https://github.com/weidai11/cryptopp/issues/652 for consistency we should add assert in all hash transformations.
Expectation is to have a good pointer and a non-0 length or a null pointer and 0-length.


noloader:
Thank you very much.

PR #655, opened 2018-05-10
Fix clang warnings in headers

MarcelRaad:
clang warns about C-style casts with `-Wold-style-cast`. It also warns about implicitly casting away const with `-Wcast-qual`. Fix both by removing unnecessary casts and converting the remaining ones to C++ casts.

Also remove an extra semicolon triggering a `-Wextra-semi` warning.


noloader:
Thanks @MarcelRaad.

PR #668, opened 2018-06-14
NaCl: add default parameter for optional random generated sk

anonimal:
>This is useful for implementations who can't afford to clobber the
existing sk buffer or who are limited by their existing interface
(e.g., they *must* re-use the privkey).

This is needed for https://github.com/monero-project/kovri/pull/909.

@noloader I know, altering the interface is heresy. I welcome your feedback.

### DIFF
--- a/naclite.h
+++ b/naclite.h
@@ -367,11 +367,12 @@ int crypto_sign_open(byte *m,word64 *mlen,const byte *sm,word64 n,const byte *pk
 /// \brief Generate a keypair for signing
 /// \param pk public key byte buffer
 /// \param sk private key byte buffer
+/// \param rnd_sk generate random private key
 /// \details crypto_sign_keypair() creates an ed25519 keypair.
 /// \returns 0 on success, non-0 otherwise
 /// \sa <A HREF="https://nacl.cr.yp.to/sign.html">NaCl crypto_sign documentation</A>
 /// \since Crypto++ 6.0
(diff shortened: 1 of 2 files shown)


anonimal:
ping @noloader

noloader:
@anonimal,

Sorry about the late reply here.

I'm not quite following what you want/need. It looks like you want a function to create a public key from a private key.

I thought libsodium provides something like that but I can't find the reference at the moment. Should we add another function that provides the functionality?

noloader:
We added a new functions called `crypto_sign_sk2pk`. Also see [Commit d8946df8eb34](https://github.com/weidai11/cryptopp/commit/d8946df8eb34).

anonimal:
Thank you so much @noloader !

Very important note: this is *not* for the Java I2P library as noted in Doxygen. This is entirely a Kovri request, and kovri has *no* java I2P interoperability (as in, we don't use their libraries in any way). They also do *not* use crypto++ in any way. We are also not associated with their project in any way outside of the open specifications that they provide.

Should I open a pull request?

anonimal:
Sorry, we're nearing a release so my stress level is to the max and I overlooked what you meant.

>no java I2P interoperability

We are interoperable, but I think it's important to note that this is a specification requirement and not a library one because A) we may continue to use this function if/when we don't use I2P (or choose not to interop with Java I2P) and B) because the function itself is a cool tool which, IMHO, can also be applied elsewhere.

PR #739, opened 2018-11-14
Misc warning fixes

bobsayshilol:
Fix some minor warnings on Clang and GCC that show up when compiling Crypto++ with `-Wall -Wextra -pedantic`.

GCC also warns about the OMP pragmas in `nbtheory.cpp`, `rw.cpp`, and `scrypt.cpp` being unknown pragmas, and guarding those with `#if defined(_OPENMP)` does get rid of the warnings, however it also makes the code messier so I opted for taking the `-Wno-unknown-pragmas` route locally rather than checking those in.

### DIFF
--- a/des.cpp
+++ b/des.cpp
@@ -284,14 +284,16 @@ namespace {
 void RawDES::RawSetKey(CipherDir dir, const byte *key)
 {
 #if (_MSC_VER >= 1600) || (__cplusplus >= 201103L)
-# define register /* Define to nothing for C++11 and above */
+# define REGISTER /* Define to nothing for C++11 and above */
+#else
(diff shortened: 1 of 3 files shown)


noloader:
> GCC also warns about the OMP pragmas ...

I consider that a GCC bug. GCC is OpenMP compliant and it knows what the macros are. The macros are being used exactly as intended.

Also see GCC Issue 66943, [GCC warns of Unknown Pragma for OpenMP, even though it support it](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66943)

noloader:
Thank you very much.

PR #752, opened 2018-12-02
Support for implementing Secret Sharing

mouse07410:
Better support for Polynomials over modulo Integer. 
Added Ring class member to avoid specifying the actual ring every time an operation is requested.

### DIFF
--- a/algebra.cpp
+++ b/algebra.cpp
@@ -12,80 +12,80 @@
 
 NAMESPACE_BEGIN(CryptoPP)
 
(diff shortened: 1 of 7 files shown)


mouse07410:
Considering the guards around that include, which make sure it is only used on Apple MacOS X, I begin to think that commit d5e27c4 is not necessary, and should be reverted.

noloader:
I'm going to close this. Feel free to resurrect it if you want to pursue it.

PR #755, opened 2018-12-04
Include with angle brackets

pirroman:
The inclusion of headers has been replaced from relative inclusion(`include "..."`) to system inclusion(`include <cryptopp/...>`). This is necessary in order to use CryptoPP in own project and the included headers of CryptoPP will not conflict with the headers of the project.


noloader:
I believe this is going to cause too many problems for folks. I'm especially concerned about the impact on Windows users. 

Perhaps you can build and install the library. That will allow you to use the `<cryptopp/file.h> in your project.

rom4ik1945:
@noloader . Before I made the PR. I supported the build on Windows (Visual Studio 2017), Linux (gcc 8.2), Clang (clang 10.0).

> Perhaps you can build and install the library. That will allow you to use the <cryptopp/file.h> in your project.

I build the library separately with conan, but without changes in this request the `include<cryptopp/file.h>` doesn't work.

irwir:
> The inclusion of headers has been replaced from relative inclusion(`include "..."`) to system inclusion(`include <cryptopp/...>`). This is necessary in order to use CryptoPP in own project and the included headers of CryptoPP will not conflict with the headers of the project.

According to C++ Standards, angle brackets should be used for searching files in "a sequence of implementation-defined places for a header".
Third-party libraries are not a part of `implementation`, therefore double quotes should be used in this case.

MarcelRaad:
I don't understand why this would be necessary. All headers are in the same directory and all compilers I know look for included files in the same path as the file containing the include statement first when using double quotes. Of course your own projects should use angle brackets. I'm using Conan too. Can you please explain when this is needed @pirroman @dmitry-zakablukov?

One problem with this will be that sometimes the include directory is called crypto++ rather than cryptopp. Can't remember where I've seen this, but there are also several find scripts looking in both directories.

noloader:
> One problem with this will be that sometimes the include directory is called crypto++ rather than cryptopp.

Red Hat and derivatives like Fedora and Scientific Linux use `cryptopp`. Debian and derivatives like Ubuntu use `crypto++`.

pirroman:
@MarcelRaad , @noloader . Thanks for your comments! The inclusion problem has solved by isolating the cryptopp in the project wrapper with inclusion of `#include<...>` instead of `#include<cryptop / ...>`. The PR is not necessary, I will decline it.

PR #765, opened 2018-12-20
add Hygon Dhyana support to enable RDRAND/RDSEED/ADX/AVX2 Feature

hygonsoc:
Enable Hygon Dhyana RDRAND/RDSEED/ADX/AVX2 Feature in cpu.cpp
all test passed


noloader:
Thanks @hygonsoc.

Ping @randombit. Botan may be interested in a change like this.

PR #766, opened 2018-12-21
Added check if VS version is 2008 or earlier to fix from 082ad86

SamuelEnglard:
A fix for Visual Studio 2008 and earlier was added in 082ad86, but the fix did not check the Visual Studio version. The fix just assumed if compiling with Visual Studio then the fix was needed. This adds a version check.


noloader:
Thanks @SamuelEnglard,

I don't recall if that was laziness or an omission on my part....

PR #771, opened 2019-01-02
Fix clang warnings

MarcelRaad:
- remove superfluous semicola to fix -Wextra-semi
- use C++ casts to fix -Wold-style-cast


noloader:
Thank you very much.

PR #774, opened 2019-01-03
Initialize "APPLE_SDK" if arm64, armv7, armv7s, armv6 or armv5 is pas…

jnavarrom:
…sed as argument


noloader:
@jnavarrom,

Sorry about the late reply and Thanks.

Damn, that is missing... Sorry about that.

PR #778, opened 2019-01-11
 Don't add the flag "DCRYPTOPP_DISABLE_ASM" twice for iPhone simulator compilation

jnavarrom:


noloader:
Thanks. That was an eyesore...

PR #809, opened 2019-02-21
config.h: avx2 depends on gcc 4.9

ffontaine:
Build of cryptopp on x86 with gcc 4.8 fails on:
chacha_avx.cpp: In function 'void CryptoPP::ChaCha_OperateKeystream_AVX2(const word32*, const byte*, CryptoPP::byte*, unsigned int)':
chacha_avx.cpp:98:85: error: '_mm256_broadcastsi128_si256' was not declared in this scope

This is due to the fact that _mm256_broadcastsi128_si256 has been added
only in gcc 4.9:
https://github.com/gcc-mirror/gcc/commit/78e8d5ffbf0ba5031b736d2c6fc6a44605047cbc

So bump CRYPTOPP_GCC_VERSION from 4.7 to 4.9 for AVX2 support

Fixes:
 - http://autobuild.buildroot.org/results/195e40b34344f773da51a3fbff9d8e76c517eed1


noloader:
Thank you very much.

Issue #812, opened 2019-02-26
Build fails on fedora30+ (CPU features miss-detected)


kwizart:
Using Fedora 30, I cannot update to cryptopp 8.0.0/8.1.0
Seems like CPU features are miss-detected as I cannot see any ASM_FLAGS been used in "some" cases.

Build attempt for f31,f29, f28
https://koji.fedoraproject.org/koji/taskinfo?taskID=33066081
https://koji.fedoraproject.org/koji/taskinfo?taskID=33066857
https://koji.fedoraproject.org/koji/taskinfo?taskID=33067570

I'm using the default GNUMake files to build using:
%make_build -f GNUmakefile \
  CXXFLAGS="-DNDEBUG %{optflags} -fPIC -DPIC" \
  LDFLAGS="%{?__global_ldflags}" \
  shared cryptest.exe

noloader:
Thanks @kwizart.

Interesting in a morbid sort of way... I use Fedora as my primary workstation, and we specifically [test using Fedora's build flags](https://github.com/weidai11/cryptopp/blob/master/cryptest.sh#L2178).

The symptom is the following taken from the [task 33066120 build.log](https://kojipkgs.fedoraproject.org//work/tasks/6120/33066120/build.log):

```
g++ -DNDEBUG -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fPIC -DPIC -c aria_simd.cpp
BUILDSTDERR: In file included from aria_simd.cpp:14:
BUILDSTDERR: /usr/lib/gcc/x86_64-redhat-linux/9/include/tmmintrin.h: In function 'void CryptoPP::ARIA_ProcessAndXorBlock_SSSE3(const byte*, CryptoPP::byte*, const byte*, CryptoPP::word32*)':
BUILDSTDERR: /usr/lib/gcc/x86_64-redhat-linux/9/include/tmmintrin.h:136:1: error: inlining failed in call to always_inline '__m128i _mm_shuffle_epi8(__m128i, __m128i)': target specific option mismatch
[...]

PR #815, opened 2019-02-28
Use TCXXFLAGS instead of CXXFLAGS

kwizart:
As done with others tests. This will avoid a miss-detection of aarch64 features
when using flags such as _FORTIFY_SOURCE that needs to be filtered for testing

This fixes https://github.com/weidai11/cryptopp/issues/812

Signed-off-by: Nicolas Chauvet <kwizart@gmail.com>

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -483,7 +483,7 @@ ifeq ($(IS_ARMV8),1)
 
   TPROG = TestPrograms/test_arm_asimd.cxx
   TOPT = -march=armv8-a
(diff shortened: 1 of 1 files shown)


noloader:
Thanks.

Oh my... That's a big oversight on my part.

noloader:
Ping @gcsideal.

Debian may want to grab this PR and [Commit 08b9e21e5a43](https://github.com/weidai11/cryptopp/commit/08b9e21e5a43). The PR applies to `GNUmakefile` and the other commit applies to `GNUmakefile-cross`.

It is not a high priority for Debian, however. Debian uses a different set of flags that are not sensitive to the `TCXXFLAGS` mis-step. Fedora encountered it because of their hardened builds and use of `_FORTIFY_SOURCE` in their `CXXFLAGS`.

PR #853, opened 2019-06-10
Compilation warning fix.

DimaStebaev:
This small pull request fixes compilation warning "Base class should be explicitly initialized in the copy constructor" in modern C++ compilers.

### DIFF
--- a/modarith.h
+++ b/modarith.h
@@ -54,7 +54,7 @@ public:
 	/// \brief Copy construct a ModularArithmetic
 	/// \param ma other ModularArithmetic
 	ModularArithmetic(const ModularArithmetic &ma)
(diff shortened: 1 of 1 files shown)


noloader:
@mouse07410,

Can you test this change to make sure it does not have unwanted side effects for your environment?

The original code used the following. It copied the modulus but nothing else:

```
ModularArithmetic(const Integer &modulus = Integer::One())
    : AbstractRing<Integer>(), m_modulus(modulus), m_result(static_cast<word>(0), modulus.reg.size()) {}
```

The default constructor was removed [recently] because of another diagnostic complaint.

mouse07410:
Unable to test right now. Will work with it tomorrow.

mouse07410:
@DimaStebaev thank you for your contribution!

PR #858, opened 2019-07-02
config: guard CRYPTOPP_SSE2_INTRIN_AVAILABLE with CRYPTOPP_DISABLE_SSE2

alonbl:
Build fails for CPU that does not support SSE2 and while CRYPTOPP_DISABLE_SSE2 is defined.


noloader:
Thank you very much.

PR #870, opened 2019-07-29
Fix ECDSA scalar multiplication leakage of bit-length.

J08nY:
This fixes the timing leakage of bit-length of nonces in ECDSA by essentially
fixing the bit-length, by using a nonce equivalent modulo the subgroup order.


noloader:
Thank you very much.

Issue #880, opened 2019-09-10
Infinite recursion in rijndael.cpp


alanbirtles:
`Rijndael::Enc::ProcessAndXorBlock` recurses infinitely under certain combinations of defines.:

- `CRYPTOPP_SSE2_ASM_AVAILABLE`  is true
- `CRYPTOPP_X64_MASM_AVAILABLE` is not defined
- `CRYPTOPP_AESNI_AVAILABLE` is not defined
- `CRYPTOPP_DISABLE_RIJNDAEL_ASM` is defined

This causes `if (HasAESNI())` to be run, if this returns true then `Rijndael::Enc::AdvancedProcessBlocks` is called. With this combination of defines `Rijndael::Enc::AdvancedProcessBlocks` just calls `Rijndael::Enc::ProcessAndXorBlock`.

This was found on OSX 10.14 with XCode 10.3 and building with the makefile with the following additional defines:
`-DCRYPTOPP_DISABBLE_SSSE3`

noloader:
Thanks @alanbirtles. Sorry about the late reply.

Yeah, I can see where this could be a problem. I guess the way to proceed is, guard the return value of `HaveXXX()` based on preprocessor macros and runtime testing (and not just runtime testing).

Let me spinup a test branch.

noloader:
@alanbirtles,

I tried to duplicate this issue on OS X, but I was not able to do so. Part of the problem may be is, I only have an old MacBook Pro with a Core2 Duo, OS X 10.8 and XCode 8.

Can you provide SSH access to the machine for testing?

I also checked-n a tentative patch at [Commit 955b2340a005](https://github.com/noloader/cryptopp/commit/955b2340a005). It is on my testing branch. Can you give it a try?

noloader:
@mouse07410,

Can you test this on your Mac?

First, duplicate the issue:

```
git clone https://github.com/weidai11/cryptopp cryptopp-aes
cd cryptopp-aes
CXXFLAGS="-DNDEBUG -g2 -O3 -DCRYPTOPP_DISABLE_SSSE3=1" make -j 4
./cryptest.exe v
```

Then, clone `git clone https://github.com/noloader/cryptopp` and perform the same test.

mouse07410:
Here are the results:
[test-out-cryptopp-aes.txt](https://github.com/weidai11/cryptopp/files/3666182/test-out-cryptopp-aes.txt)
[...]

PR #886, opened 2019-09-29
Fix Rijndael recursion overflow when -DCRYPTOPP_DISABLE_SSSE3 (GH #880)

noloader:
Fix Rijndael recursion overflow when -DCRYPTOPP_DISABLE_SSSE3 (GH #880)


noloader:
Thanks @mouse07410 and @alanbirtles.

PR #888, opened 2019-09-30
blowfish: fix -Wundefined-var-template clang warning

rzvncj:
By adding explicit instantiation declarations.


noloader:
Thank you very much.

rzvncj:
No problem, sorry for unwittingly causing the warning previously.

PR #893, opened 2019-10-15
Fix setenv-android.sh and cryptest-android.sh for NDK r19

noloader:
`setenv-android.sh` and `cryptest-android.sh` are current for NDK r19 again.

NDK r17 and below is broken. We'll try to get to it next.


noloader:
Ping @Skycoder42. I think he has suffered these files as much as I have...

PR #897, opened 2019-10-17
config_asm.h: Allow clang-tooling

t-b:
When using cryptopp in your own projects and using clang tooling (e.g.
clang-tidy) for inspecting your project, this currently errors out due to the #error
directive when done on Windows with MS Visual Studio.

The #error directive here is meant for code execution and not for code
inspection.

So we now don't error out  when __clang_analyzer__ is defined. That
macro was introduced into clang tooling in
https://reviews.llvm.org/D46325.


noloader:
Thanks @t-b.

PR #925, opened 2020-01-22
Add vcpkg installation instructions

PhoebeHui:
Cryptopp is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build cryptopp, ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for cryptopp and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/Install.txt
+++ b/Install.txt
@@ -82,6 +82,19 @@ Crypto++ does not enagage Specter remediations at this time. You can build with 
 
     CXXFLAGS="-DNDEBUG -g2 -O3 -mfunction-return=thunk -mindirect-branch=thunk" make
 
+BUILDING WITH VCPKG
+-------------------
+
+You can download and install cryptopp using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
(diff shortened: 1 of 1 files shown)


noloader:
Thank you very much.

PhoebeHui:
Thanks for your help!

noloader:
@PhoebeHui,

Another library you may be interested in providing (if you have not done so) is [Botan](https://github.com/randombit/botan). It is a C++ security and TLS library by Jack Lloyd and his team. It is a hidden gem on the web.

Ping @randombit.

PhoebeHui:
@noloader, thanks for mentioning this, the instructions has been added to [wiki ](https://github.com/randombit/botan/wiki/Distros).

PR #928, opened 2020-01-28
files.cpp Reset file ptr rather than release on init

bgura:
It looks like there was a possible memleak here or perhaps i'm misunderstanding the function of this class. 

Multiple calls to IsolatedInitialize (public function) would call release on the m_file member which removes the ownership of the underlying ostream from m_file and then that instance is lost and never deleted.

It appears m_file is owned by the FileStore class. On Destruction of FileStore, it will free the resource stored in m_file. File store should also be deleting this memory before it assigns a new instance, no?

### DIFF
--- a/files.cpp
+++ b/files.cpp
@@ -23,7 +23,7 @@ void FileStore::StoreInitialize(const NameValuePairs &parameters)
 {
 	m_waiting = false;
 	m_stream = NULLPTR;
-	m_file.release();
+	m_file.reset(nullptr);
 
 	const char *fileName = NULLPTR;
 #if defined(CRYPTOPP_UNIX_AVAILABLE) || _MSC_VER >= 1400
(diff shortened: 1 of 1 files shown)


noloader:
The `FileSource` and `FileSInk` take a reference, not a pointer:

```
FileStore(std::istream &in) : m_stream(NULLPTR), m_space(NULLPTR), m_len(0), m_waiting(0)
    {StoreInitialize(MakeParameters(Name::InputStreamPointer(), &in));}
```

Because they take a reference, the caller owns the object. The caller is responsible for cleaning up the `std::istream` pointer.

If they had taken a pointer, then the `FileSource` and `FileSInk` would own the `std::istream` and they would delete it.

Also see [Pipelines | Ownership](https://www.cryptopp.com/wiki/Pipelining#Ownership) on the Crypto++ wiki.

PR #929, opened 2020-02-07
Add asm code to enable solution Borland/Embarcadero and Issue 500

dafajopa:
returns from cpuid call, OK.
Ok for _WIN64 and _WIN32 (as before) and __APPLE__.

### DIFF
--- a/cpu.cpp
+++ b/cpu.cpp
@@ -172,6 +172,23 @@ bool CpuId(word32 func, word32 subfunc, word32 output[4])
 		// Borland/Embarcadero and Issue 500
 		// Local variables for cpuid output
 		word32 a, b, c, d;
+#if defined(_WIN64)
+		__asm
+		{
+			push rdi
+			push rbx
+			mov eax, func
+			mov ecx, subfunc
+			cpuid
+			mov edi, ebx
+			pop rbx
+			mov b, edi
+			mov a, eax
+			mov c, ecx
+			mov d, edx
+			pop rdi
+		}
+#else
 		__asm
 		{
 			mov eax, func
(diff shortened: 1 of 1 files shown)


noloader:
@dafajopa,

Forgive my ignorance... All the cpuid code I have seen uses 32-bit registers. Why doe Borland need 64-bit code?

noloader:
@dafajopa,

After thinking about this... Do we need a `CPUID64` function for Microsoft platforms, similar to `XGETBV64`? `XGETBV64` is provided in a [MASM file](https://github.com/weidai11/cryptopp/blob/master/x64dll.asm#L1967) because we can't inline it.

dafajopa:
It is possible to move my _WIN64 code to another method (I dont know your organisations).

About the _WIN64 asm code : It's seems the registers in _WIN64 are not used in the same way as in _WIN32.  So I push the 'rbx', because I receive data in 'ebx' from cpuid and it is needed by the "move a/b/c/d , ...". I push 'rdi' because I use 'edi' to transfert data from 'ebx' to 'a'.

About the catch(...) : I don't like it to, but it's the only way I found to enable the code to work on macOS. I haven't put the specific behavior for Android or iOS, but I will do it later.

noloader:
@dafajopa,

I just broke your pull request... Sorry about that.

I added the `CPUID64` function for WIN64. See if you can get Borland into the code path around `cpu.cpp:150`. Borland uses Clang (IIRC) and it may have the Microsoft intrinsics.

```
#if _MSC_VER >= 1600

inline bool CpuId(word32 func, word32 subfunc, word32 output[4])
{
	__cpuidex((int *)output, func, subfunc);
	return true;
}

# elif ...
```

noloader:
@dafajopa,

I just checked in a cleanup of the `CpuId` function. It should be a little easier to follow the conditions. Try to get Borland into the MSVC paths on Windows, and into the Unix and Linux path on OS X.

Also see [Commit dd67e1eee46b](https://github.com/weidai11/cryptopp/commit/dd67e1eee46b).

dafajopa:
OK, I will try with your modificatoin. I will give you a feed back.
I forget, when I compile the library I need to link the library "compiler_rt.a", but everything OK !

You can download a **C++Builder Community Edition** for free.
https://www.embarcadero.com/fr/products/cbuilder/starter

Best regards.

PR #942, opened 2020-03-31
Fix compile warnings

akovachev:
The double semicolons are generating compile errors which breaks builds that have "compile warnings as errors".


noloader:
thanks @akovachev .

PR #949, opened 2020-05-14
Better shell compatibility in setenv-ios.sh

arhoads:
For better shell compatibility I changed the double "==" to single "=".

The zsh shell I use on my Mac machine errors on the double equals when I run `source setenv-ios.sh`.

### DIFF
--- a/TestScripts/setenv-ios.sh
+++ b/TestScripts/setenv-ios.sh
@@ -47,23 +47,23 @@ unset IOS_SYSROOT
 #####    Small Fixups, if needed    #####
 #########################################
 
-if [[ "$IOS_SDK" == "iPhone" ]]; then
+if [[ "$IOS_SDK" = "iPhone" ]]; then
     IOS_SDK=iPhoneOS
 fi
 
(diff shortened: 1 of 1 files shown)


noloader:
@arhoads,

If single `=` is used in Bash, then we should use a single bracket `[` or `]`. I.e.,

```
if [ "$IOS_SDK" = "iPhone" ]; then
    IOS_SDK=iPhoneOS
fi
```

Is that OK with zsh? If so, can you make the change in all of the setenv-*.sh scripts?

arhoads:
@noloader My mistake. I was having the problem with 8.2.0, but this isn't a problem in master. 

zsh errored on `if [ "$APPLE_SDK" == "iPhoneSimulator" ]; then`.

zsh doesn't mind `==` with `[[ ]]`. The new script is fine: 
`if [[ "$IOS_SDK" == "iPhoneSimulator" || "$IOS_SDK" == "AppleTVSimulator" || "$IOS_SDK" == "WatchSimulator" ]]; then`.

Sorry about that!

PR #961, opened 2020-07-29
improve runtime exception messages

Arzaghi:
Improve runtime exception messages to guide the user to choose correct value for each invalid parameter

### DIFF
--- a/chacha.cpp
+++ b/chacha.cpp
@@ -326,7 +326,7 @@ void ChaCha_Policy::CipherSetKey(const NameValuePairs &params, const byte *key, 
     // Use previous rounds as the default value
     int rounds = params.GetIntValueWithDefault(Name::Rounds(), m_rounds);
     if (rounds != 20 && rounds != 12 && rounds != 8)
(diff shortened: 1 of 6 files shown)


noloader:
@mouse07410,

Any thoughts on this?

Arzaghi:
> @mouse07410,
> 
> Any thoughts on this?

I decided to use this amazing library but finding the correct values for different parameters of different algorithms was quite difficult. so I decided to improve runtime exception messages. so anybody on higher level of applications can use these messages to find the reasons of the problem and how to choose correct values.

expnkx:
> > @mouse07410,
> > Any thoughts on this?
> 
> I decided to use this amazing library but finding the correct values for different parameters of different algorithms was quite difficult. so I decided to improve runtime exception messages. so anybody on higher level of applications can use these messages to find the reasons of the problem and how to choose correct values.

I think these issues will only get systematically solved after herb sutter's deterministic exceptions got added into C++.

noloader:
I think we are going to pass on the PR. The problem is, we risk breaking scripts that call Crypto++ based programs. The original messages have been present for about 30 years. If the message changes it could have negative impact on others.

Issue #966, opened 2020-09-14
memset_z function is optimized out by some compilers


jm1hey:
## Background
The memset_z function is defined in misc.h, and is intended to act as a drop-in replacement for memset (provided in the cstring header of the standard library). The important feature of memset_z is that calls to memset_z should never be optimized out by a compiler. This is due to its implementation, which involves the use of the 'volatile' keyword.

Normally, if a compiler sees that after a call such as "memset(foo, 0, foo_size)" foo is never read or written, the compiler can ignore the call in order to improve execution speed. However, if the contents of foo are sensitive and should be cleared from memory before the memory is deallocated (e.g. foo contains a secret key), then it is desirable to ensure the compiler doesn't ignore the call.

Compiler optimizations can be configured at will; they are usually enabled in a build system's release configuration and disabled in a debug configuration.

In Crypto++ 8.2.0, memset_z is called in misc.h and in secblock.h .

## Issue
In Crypto++ 8.2.0, memset_z can, in fact, be optimized out during compilation.

Consider the following test code, which uses an exact copy of memset_z as written in misc.h:

	#include <cstring>
	#include <iostream>

	inline void * memset_z(void *ptr, int value, size_t num)
	{
		// avoid extranous warning on GCC 4.3.2 Ubuntu 8.10
	#if CRYPTOPP_GCC_VERSION >= 30001
		if (__builtin_constant_p(num) && num==0)
			return ptr;
	#endif
		volatile void* x = memset(ptr, value, num);
		return const_cast<void*>(x);
	}

	int main(void)
	{
		char str[10] = "abc123xyz";
		std::cout << str << std::endl;
		memset_z(str, 0, 10);
		return 0;
	}

I compiled this code using Visual Studio 2019 and CMake with the following compilers:
* MSVC 14.26
* Clang 10.0.0
* Mingw64 g++ 10.2.0

In each case I used the 'release with debug information' configuration, and placed a breakpoint at the return statement in main(). At that point, watching 'str' shows that it still contains "abc123xyz" whereas it should be an empty string. Of course, using a debug configuration (optimizations disabled) results in the correct behavior.

## Proposed solution
Either of the following alternative implementations of memset_z results in the correct behavior in all of the compilers that I tried:

	typedef void* (*memset_type)(void*, int, size_t);
	volatile memset_type memset_z = memset;
or

	inline void * memset_z(void *ptr, int value, size_t num)
	{
[...]

PR #967, opened 2020-09-15
Make memset_z less vulnerable to optimization

jm1hey:
Change memset_z to use a loop with a volatile pointer so that it is more resistant to being optimized out by compilers

This is intended to address the following issue:
weidai11/cryptopp#966

### DIFF
--- a/misc.h
+++ b/misc.h
@@ -610,8 +610,8 @@ inline void vec_swap(T& a, T& b)
 /// \param ptr pointer to the memory block being written
 /// \param value the integer value to write for each byte
 /// \param num the size of the source memory block, in bytes
-/// \details Internally the function calls memset with the value value, and receives the
-///  return value from memset as a <tt>volatile</tt> pointer.
+/// \details Internally the function loops through the block, setting each byte
+///  to the given value using indirection by way of a <tt>volatile</tt> pointer.
 inline void * memset_z(void *ptr, int value, size_t num)
 {
 // avoid extranous warning on GCC 4.3.2 Ubuntu 8.10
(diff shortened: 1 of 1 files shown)


noloader:
Thanks @jm1hey,

I think approach is, drop `volatile` altogether. Looking at the original [`memset_z` from Crypto++ 5.6.2](https://github.com/weidai11/cryptopp/blob/789f81f048c9), there was no `volatile`.

Dropping the `volatile`  should be safe in this instance. `memset_z` is used to zero a freshly allocated block, and _not_ wipe a block during destruction.

I vaguely remember adding `volatile` several years ago while working on another problem. In hindsight it was not needed. And adding it was a problem if it is causing users to waste time with it.

expnkx:
Hi noloader. I know you do not like me.
https://www.usenix.org/sites/default/files/conference/protected-files/usenixsecurity17_slides_zhaomo_yang.pdf

This security paper said the correct method should be:

```cpp
#if defined(_MSC_VER)
void __stdcall RtlSecureZeroMemory(void*,std::size_t) noexcept;
#endif

inline void secure_clear(void* data,std::size_t size) noexcept
{
#if defined(_MSC_VER)
	RtlSecureZeroMemory(data, size);
#else
	std::memset(data,0,size);
	__asm__ __volatile__("" ::"r"(data): "memory");
#endif
}

```

jm1hey:
> Thanks @jm1hey,
> 
> I think approach is, drop `volatile` altogether. Looking at the original [`memset_z` from Crypto++ 5.6.2](https://github.com/weidai11/cryptopp/blob/789f81f048c9), there was no `volatile`.
> 
> Dropping the `volatile` should be safe in this instance. `memset_z` is used to zero a freshly allocated block, and _not_ wipe a block during destruction.
> 
> I vaguely remember adding `volatile` several years ago while working on another problem. In hindsight it was not needed. And adding it was a problem if it is causing users to waste time with it.

You are right. On closer inspection of functions that call memset_z, there is no reason for it to be protected against optimization. I was going by the function documentation, which states that it "attempts to survive optimizations." Closing.

PR #968, opened 2020-09-19
Clear failbit after failed istream::seekg()

tanzislam:
`std::istream` does not generally need to support seek-to-end behavior; it depends on the underlying `std::streambuf` implementation. If it fails the stream will set `failbit`, after which any further operation would fail until it is cleared.

Reference:
 - https://en.cppreference.com/w/cpp/io/basic_istream/seekg
 - https://en.cppreference.com/w/cpp/io/basic_streambuf/pubseekoff

This issue surfaced in [my use case](https://github.com/tanzislam/cryptopals/issues/32) after 04dee2a46, as I am using several `std::streambuf`-derived classes as streaming codecs and they do not support a seek-to-end functionality.

### DIFF
--- a/files.cpp
+++ b/files.cpp
@@ -67,6 +67,7 @@ lword FileStore::MaxRetrievable() const
 
 	std::streampos current = m_stream->tellg();
 	std::streampos end = m_stream->seekg(0, std::ios::end).tellg();
+	m_stream->clear();
 	m_stream->seekg(current);
 	return end-current;
 }


noloader:
Thank you very much.

noloader:
@tanzislam,

I'm concerned about `return end-current` when `end` is `streambuf(-1)` due to an error.

In general, the library understands `LWORD_MAX` or `-1` to mean "process all bytes". However, if 10 bytes have been read, then the return value is not `LWORD_MAX` or `-1` (it is `LWORD_MAX-10`), so the library misses the special value and does dumb things like trying to allocate a massive `LWORD_MAX-10` bytes. That was the problem in GH # 962.

Would the following be more robust?

```
$ git diff
diff --git a/files.cpp b/files.cpp
index 2d05e03f..128d1bd3 100644
--- a/files.cpp
+++ b/files.cpp
@@ -62,13 +62,20 @@ void FileStore::StoreInitialize(const NameValuePairs &parameters)

 lword FileStore::MaxRetrievable() const
 {
-       if (!m_stream)
+       if (!m_stream || m_stream->eof())
                return 0;

+       // Clear error bits due to seekg(). Also see
+       // https://github.com/weidai11/cryptopp/pull/968
        std::streampos current = m_stream->tellg();
        std::streampos end = m_stream->seekg(0, std::ios::end).tellg();
        m_stream->clear();
        m_stream->seekg(current);
+
+       // Return max for a non-seekable stream
+       if (end == std::streampos(-1))
+               return LWORD_MAX;
+
        return end-current;
 }
```

tanzislam:
Yes, that return expression looks reasonable to me.

BTW the current form of the code assumes that seek-to-end may be unsupported, but the seek to current position (after the `clear()`) is supported. It's only so because I was able to [twist my `streambuf` implementations into being a successful no-op](tanzislam/cryptopals@a600466) when told to seek to current position. However, truly un-seek-able streams like `std::cin` might not implement even that much behavior, so the code as-is might still leave `failbit` set. So I'd recommend putting in another `clear()` after the second `seekg()`, before returning.

expnkx:
hello. iostream also supports throwing exceptions. I think you probably need to guard your code with try catch?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

noloader:
@expnkx,

> iostream also supports throwing exceptions. I think you probably need to guard your code with try catch?

Yeah, I was not sure how to handle that. I punted and left it to the caller (by doing nothing).

One way is a try/catch. Another is to save the exception mask, 0 the mask, then restore it. I think the 0-mask would be the better approach to avoid the exception given failures are also indicated with `std::std::streampos(-1)`.

PR #990, opened 2020-12-27
Build with Oracle Developer Studio 12.6 in 64-bit

tanzislam:
- The `_mm256_loadu_si256()` intrinsic is declared with a pointer to non-`const` input. Apply `const_cast` accordingly.

- The `XGETBV` instruction isn't recognized, so apply the same workaround as for old GCC versions.
  https://docs.oracle.com/cd/E53394_01/html/E54851/gnyen.html

- Only AT&T syntax is supported; there is no directive to switch Intel syntax. Therefore, disable mixed-ASM code paths.


noloader:
Thanks @tanzislam .

> The _mm256_loadu_si256() intrinsic is declared with a pointer to non-const input. Apply const_cast accordingly.

Man, is that compiler bug still present? I wish the Oracle folks would fix it already...

tanzislam:
They may have fixed it, but only in the $$$$$ version. 🙂

PR #1000, opened 2021-01-24
Fix spelling issue

Hrxn:
`Determinsitic` -> `Deterministic`


noloader:
Thank you very much. It is amazon how bad my spelling is. You would not know I actually went to college. Derp...

Hrxn:
Yeah it happens.. 👍

PR #1017, opened 2021-03-11
Support querying Windows 10 ARM64 CPU features.

wyattoday:
This is in addition to the existing probing that exists.

### DIFF
--- a/cpu.cpp
+++ b/cpu.cpp
@@ -20,6 +20,12 @@
 # include <immintrin.h>
 #endif
 
+// For IsProcessorFeaturePresent() on Microsoft Arm64 platform
+#if defined(_WIN32) && defined(_M_ARM64)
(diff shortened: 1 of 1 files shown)


noloader:
Thank you very much @wyattoday.

While you are at it, could you look at @randombit's [Botan](https://github.com/randombit/botan)? Botan may need similar changes.

noloader:
@wyattoday,

I removed SHA512 from the list. SHA3 and SHA512 are ARMv8.2-a. They will need a separate enumerated value in MSDN.

wyattoday:
Sounds good. Hopefully MS will keep up with newer ARM processors and update their headers.

Regarding Botan, I have no experience with that project and wouldn't feel comfortable poking around in something I don't use.

PR #1027, opened 2021-04-21
Allow building with Intel oneAPI DPC++/C++ compiler

tanzislam:
This new Clang-based compiler (`icx`) apparently #defines `__CLANG__`, even
though it is not documented.
https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compiler-reference/macros/use-predefined-macros-for-intel-compilers.html

### DIFF
--- a/config_os.h
+++ b/config_os.h
@@ -29,7 +29,7 @@
 // https://www.cryptopp.com/wiki/Release_Process#Self_Tests
 // Some relevant bug reports can be found at:
 // * Clang: http://github.com/weidai11/cryptopp/issues/147
-#if (defined(_MSC_VER) && defined(__clang__) && !(defined( __clang_analyzer__)))
+#if (defined(_MSC_VER) && defined(__clang__) && !(defined( __clang_analyzer__)) && !defined(__INTEL_LLVM_COMPILER))
 # error: "Unsupported configuration"
 #endif
 


noloader:
Thanks @tanzislam.

Do we need to make a change in [config_ver.h](https://github.com/weidai11/cryptopp/blob/master/config_ver.h)? Will we mis-identify ICPC as Clang?

tanzislam:
`icpc` (i.e. the classic Intel C++ compiler) is documented to have a different identification macro `__INTEL_COMPILER`, so it should behave the same way as before. (Haven't tested, as I don't have a license to it.)

noloader:
Thanks again @tanzislam.

Regarding `CRYPTOPP_DISABLE_SHANI`, do you know why we are misdetecting the feature?

Didi I simply get it wrong at https://github.com/weidai11/cryptopp/blob/master/config_asm.h#L170 ?

Here's the block we use:

```c++
#if !defined(CRYPTOPP_DISABLE_SHANI) && defined(CRYPTOPP_SSE42_AVAILABLE) && \
	(defined(__SHA__) || (CRYPTOPP_MSC_VERSION >= 1900) || (__SUNPRO_CC >= 0x5160) || \
	(CRYPTOPP_GCC_VERSION >= 40900) || (__INTEL_COMPILER >= 1300) || \
	(CRYPTOPP_LLVM_CLANG_VERSION >= 30400) || (CRYPTOPP_APPLE_CLANG_VERSION >= 50100))
	#define CRYPTOPP_SHANI_AVAILABLE 1
#endif
```

noloader:
@tanzislam,

I think I used the wrong version of ICPC for SHA support. I believe v13.0 was wrong. It looks like it is closer to v16.0 or v16.1.

I changed to version to 16.0 at [Commit b206071aa7cc](https://github.com/weidai11/cryptopp/commit/b206071aa7cc).

tanzislam:
I can't say whether the `(__INTEL_COMPILER >= 1300)` check was right or not, as it was intended for the classic Intel compiler (`icpc`/`icl`) whereas I am using the new Intel oneAPI compiler (`icx`) that doesn't define `__INTEL_COMPILER` at all.

I think my build with `icx` managed to get `CRYPTOPP_SHANI_AVAILABLE` through `CRYPTOPP_MSC_VERSION` (which depends on `_MSC_VER`) or `CRYPTOPP_LLVM_CLANG_VERSION` (which depends on `__clang__`). If `icx` doesn't implement SHA intrinsics yet, then perhaps `__INTEL_LLVM_COMPILER` needs to be excluded in that logic.

noloader:
Thanks again @tanzislam,

From Intel's {ICX Developer Guide](https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top.html) they say they support Intel intrinsics. That should include SHA.

If you have an Intel account, maybe you can file a bug report. You can use this as your reproducer: [TestPrograms/test_x86_sha.cpp](https://github.com/weidai11/cryptopp/blob/master/TestPrograms/test_x86_sha.cpp). Maybe something like:

```c
icx.exe TestPrograms/test_x86_sha.cpp /out:test.exe
```

PR #1029, opened 2021-04-25
Implement SIMD checks for FreeBSD/powerpc*

pkubaj:
FreeBSD uses elf_aux_info() instead of getauxval().

### DIFF
--- a/cpu.cpp
+++ b/cpu.cpp
@@ -1208,6 +1208,11 @@ inline bool CPU_QueryAltivec()
 	unsigned int unused, arch;
 	GetAppleMachineInfo(unused, unused, arch);
 	return arch == AppleMachineInfo::PowerMac;
+#elif defined(__FreeBSD__) && defined(PPC_FEATURE_HAS_ALTIVEC)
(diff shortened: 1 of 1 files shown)


noloader:
Thank you very much.

We are probably missing a few FreeBSD specific checks.

noloader:
Thanks again @pkubaj,

It looks like we are catching compile failures on FreeBSD 12 and 13. See https://github.com/noloader/cryptopp/runs/2437036069.

It looks like a FreeBSD bug to me. Including their headers causes the compile failure (https://github.com/weidai11/cryptopp/blob/master/cpu.cpp#L60):

```c++
#if defined(__FreeBSD__)
# include <machine/cpu.h>
# include <sys/auxv.h>
# include <sys/elf_common.h>
#endif
```

Would you be able to let the FreeBSD folks know?

noloader:
I ended up removing the problem hear, which was `machine/cpu.h`. Also see [Commit 2619dbec0bfc](https://github.com/weidai11/cryptopp/commit/2619dbec0bfc).

PR #1030, opened 2021-04-25
Correct crypto macro on FreeBSD

pkubaj:
It's actually named `PPC_FEATURE2_HAS_VEC_CRYPTO` on FreeBSD. I noticed it right after sending my original PR, but you were too quick for me to fix it :)


noloader:
No problems. I found your other patch for NSS when verifying the previous PR.

PR #1035, opened 2021-04-27
chore: add .bat file

AJTL:
auto copy .h to cryptopph
 not include resource.h


noloader:
No thanks. You should keep this as a local patch.

PR #1039, opened 2021-05-17
Add CRYPTOPP_SUPPORT_CLANG_CL option.

lemourin:
This allows to use clang-cl to compile cryptopp by defining a macro at compile time.


noloader:
I don't believe we need this since [Commit 26a819d5c26c](https://github.com/weidai11/cryptopp/commit/26a819d5c26c).

PR #1043, opened 2021-06-09
Fix issue with VS runtime checks for data truncation

ejl103:
When running VS with the 'c' runtime check on for data truncation (see https://docs.microsoft.com/en-us/cpp/build/reference/rtc-run-time-error-checks?view=msvc-160)

This code will throw errors because the 64 bit ptr gets truncated to a 32bit value. Normally I'd disable this using the pragma specifically for this code, but sadly that doesn't work for template functions, so I have come up with a way to get the appropriate masking type for any template combinations and using the recommended way to silence the error without impacting the quality of the generated code.

This code could be more concise in modern C++ but it seems the project expects to support C++03 so this should work fine there.

All tests passed locally for me on Windows (some asserts but verified they were the same before my changes)

### DIFF
--- a/misc.h
+++ b/misc.h
@@ -1113,6 +1113,47 @@ inline T1 SaturatingSubtract1(const T1 &a, const T2 &b)
 	return T1((a > b) ? (a - b) : 1);
 }
 
+/// \brief Get the appropriate mask to do a truncating cast
+template <class T, size_t SIZE = sizeof(T)>
+struct TruncatingMask;
+
+template <class T>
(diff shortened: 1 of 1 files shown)


noloader:
Sorry about the delay in this. I often miss PR's that are sitting in queue.

Would this work for you:

```
return T2(a & SaturatingSubtract(b,1U));
```

The `SaturatingSubtract` creates the mask, and its type is `T2`.

ejl103:
Yes I think is fine, sorry I missed the simpler solution there!

PR #1062, opened 2021-07-20
Load msvcrt explicitly when not loaded

Algebro7:
EVE Online started shipping with the cryptopp DLL recently, which was crashing when being ran with Wine on Linux. After a bunch of debugging, I found out that `dll.cpp` was assuming `msvcrt.dll` was already loaded, which failed since the wine system binaries are not linked with msvcrt (only `ucrtbase` as of wine 5.9). This patch loads `msvcrt.dll` if module handles aren't found for `msvcrt` and `msvcrtd` before attempting to set the pointers.

### DIFF
--- a/dll.cpp
+++ b/dll.cpp
@@ -119,8 +119,11 @@ static void SetNewAndDeleteFunctionPointers()
 	// try getting these directly using mangled names of new and delete operators
 
 	hModule = GetModuleHandle("msvcrtd");
-	if (!hModule)
+	if (!hModule) 
 		hModule = GetModuleHandle("msvcrt");
+	if (!hModule)
+		hModule = LoadLibrary("msvcrt.dll");
+	
 	if (hModule)
 	{
 		// 32-bit versions
@@ -135,7 +138,7 @@ static void SetNewAndDeleteFunctionPointers()
 		if (s_pNew && s_pDelete)
 			return;
 	}
-
+		
 	OutputDebugStringA("Crypto++ DLL was not able to obtain new and delete function pointers.\n");
 	throw 0;
 }


Algebro7:
Relevant wine HQ ticket: https://bugs.winehq.org/show_bug.cgi?id=51465

noloader:
@Algebro7,

Forgive my ignorance... Is EVE Online an open source project? I'm trying to find their sources to see exactly what they are doing.

Stepping back to 10,000 feet, here the issue. The [Crypto++ DLL](https://www.cryptopp.com/wiki/FIPS_DLL) is a specialized DLL. Its intended purpose was to provide validated cryptography for apps that needed them on Windows. Crypto++ is now on the historical validation list so the library cannot be used anymore. The FIPS DLL is planned for removal very soon.

When the FIPS DLL goes away the library will still provide cryptography. It just won't be the formerly validatedcryptography. But that means apps that depend on a Crypto++ DLL will need to (1) use the static lib on Windows, or (2) provide/create their own DLL,. Both are equally easy.

So I would like to speak with the EVO dev team and get them to a place where they don't need to worry about what happens to the FIPS DLL. I'm happy to work with them so it is an easy process.

As for the PR, I can merge it. But you are just fixing a symptom. It s not fixing the underlying problems caused by the FIPS DLL.

Algebro7:
Thanks for the details! I've mentioned this on the eve forum thread [here](https://forums.eveonline.com/t/new-launcher-new-linux-error/319782/199), but I haven't been able to directly get in contact with the developers of the Eve Launcher (CCP_Cemetary in that thread). The fact that removal of the DLL will impact Windows users is something they will probably want to take action on.

noloader:
@Algebro7,

Yeah, the only contact I could find is info@ccpgames.com. This is typical when dealing with some companies. We want to help but we can't find a good point of contact.

Algebro7:
The producer for the Launcher that is affected by this mentioned he would respond to support tickets--I've opened one and linked to your comment above. I'll let you know if they get in touch. Thanks again for your help!

Algebro7:
I was not able to reach the person in charge of the Launcher (the support employee would not escalate the ticket that far), they basically said they're reviewed this thread but they're not planning to do anything about it or reach out to you to resolve it. Given that, would it be possible to merge this PR at some point to hold us over until the DLL is removed? I know even if we do, it would require them to update to the next stable release which they probably won't bother to do, but at least it would be some forward progress.

Algebro7:
Just a quick update @noloader , the wine devs helped us troubleshoot and [submitted a patch](https://source.winehq.org/patches/data/209969) that fixes the specific problem with Wine not having msvcrt.dll loaded. So we are probably good to close this if you'd prefer, unless you want to merge it until the DLL is removed.

noloader:
@Algebro7,

I was looking at the message board thread at https://forums.eveonline.com/t/new-launcher-new-linux-error/319782/191 . I see you guys were discussing <tt>g_powerUpSelfTestStatus</tt> and friends.

The details of what happens is discussed in the Crypto++ wiki at [FIPS_DLL | Implementation_Details](https://www.cryptopp.com/wiki/FIPS_DLL#Implementation_Details). The easiest hack might be to remove calls to `DoPowerUpSelfTest`:

```
$ grep 'DoPowerUpSelfTest' *.h *.cpp *.proto
fips140.h:CRYPTOPP_DLL void CRYPTOPP_API DoPowerUpSelfTest(const char *moduleFilename, const byte *expectedModuleMac);
fipstest.cpp:void DoPowerUpSelfTest(const char *moduleFilename, const byte *expectedModuleMac)
fipstest.cpp:   CryptoPP::DoPowerUpSelfTest(NULLPTR, s_moduleMac);
```

And set `g_powerUpSelfTestStatus` to `POWER_UP_SELF_TEST_PASSED`.

```
$ grep 'g_powerUpSelfTestStatus =' *.cpp
fips140.cpp:PowerUpSelfTestStatus g_powerUpSelfTestStatus = POWER_UP_SELF_TEST_NOT_DONE;
fips140.cpp:    g_powerUpSelfTestStatus = POWER_UP_SELF_TEST_FAILED;
fipstest.cpp:   g_powerUpSelfTestStatus = POWER_UP_SELF_TEST_NOT_DONE;
fipstest.cpp:           g_powerUpSelfTestStatus = POWER_UP_SELF_TEST_FAILED;
fipstest.cpp:   g_powerUpSelfTestStatus = POWER_UP_SELF_TEST_PASSED;
```

noloader:
@Algebro7,

I just posted this to the mailing list: [Removal of the FIPS DLL at next Crypto++ release](https://groups.google.com/g/cryptopp-users/c/2GQjimCcACw).

Algebro7:
Thanks for looking into it! With the powerupselftest issue, I'm thinking it could've been related to running the program in a debugger, because I stopped being able to reproduce it later (and the game runs fine with the upstream wine patch that imports msvcrt in advapi32.dll). So fortunately that problem seems to have gone away.

As for removing the DLL, sounds fine to me. I tried to give CCP a heads up and told them the cryptopp maintainers offered to help them transition and they wanted nothing to do with it. So that'll be their problem to deal with when it happens.

PR #1083, opened 2021-11-01
Update .gitignore

jamiehankins:
This ignores the .vs directory at the root. This is used by Visual Studio to save user configuration that should not be in source control.


noloader:
Thanks.

PR #1099, opened 2022-01-04
Fix typos

rex4539:


noloader:
Thanks @rex4539.

Not all of them are me. I believe this was Wei:

```c++
throw InvalidArgument("Mash: output legnth too large");
```

Man, my spelling is getting worse as I get older....

PR #1119, opened 2022-04-29
Fix setenv-android.sh for NDK version 22 and higher

radioburst:
Older NDKs had three linkers "ld", "ld.bfd" and "ld.gold". The current script is always using the file without any post fix. But it seems the "ld" was removed since NDK 22 (I used 22.1.7171670) and now only the two executables with bfd and gold post fixes are available. Since the script can't handle this I added a new var "ANDROID_LD" which lets the user decide which linker should be used. If not set it will default to the old behavior.

e.q.:
`ANDROID_API="30" ANDROID_CPU="arm64-v8a" ANDROID_LD="gold" source TestScripts/setenv-android.sh`

### DIFF
--- a/TestScripts/setenv-android.sh
+++ b/TestScripts/setenv-android.sh
@@ -371,7 +371,7 @@ fi
 
 # Error checking
 if [ ! -e "${ANDROID_TOOLCHAIN}/$LD" ]; then
-    echo "ERROR: Failed to find Android ld. Please edit this script. When using NDK 22 or hihger make sure to set ANDROID_LD! (bfd, gold)"
+    echo "ERROR: Failed to find Android ld. Please edit this script. When using NDK 22 or higher make sure to set ANDROID_LD! (bfd, gold)"
     [ "$0" = "${BASH_SOURCE[0]}" ] && exit 1 || return 1
 fi
 
--- a/TestScripts/setenv-android.sh
+++ b/TestScripts/setenv-android.sh
@@ -202,7 +202,11 @@ case "$ANDROID_CPU" in
   armv7*|armeabi*)
     CC="armv7a-linux-androideabi${ANDROID_API}-clang"
     CXX="armv7a-linux-androideabi${ANDROID_API}-clang++"
-    LD="arm-linux-androideabi-ld"
+    if [ -z "${ANDROID_LD}" ]; then
+        LD="arm-linux-androideabi-ld"
+    else
+        LD="arm-linux-androideabi-ld.${ANDROID_LD}"
+    fi
     AS="arm-linux-androideabi-as"
     AR="arm-linux-androideabi-ar"
     RANLIB="arm-linux-androideabi-ranlib"
@@ -228,7 +232,11 @@ case "$ANDROID_CPU" in
   armv8*|aarch64|arm64*)
     CC="aarch64-linux-android${ANDROID_API}-clang"
     CXX="aarch64-linux-android${ANDROID_API}-clang++"
-    LD="aarch64-linux-android-ld"
+    if [ -z "${ANDROID_LD}" ]; then
+        LD="aarch64-linux-android-ld"
+    else
+        LD="aarch64-linux-android-ld.${ANDROID_LD}"
+    fi
     AS="aarch64-linux-android-as"
     AR="aarch64-linux-android-ar"
     RANLIB="aarch64-linux-android-ranlib"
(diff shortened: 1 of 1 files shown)


Susangriffis (approved):

TestScripts/setenv-android.sh
```diff
@@ -355,7 +371,7 @@ fi
 
 # Error checking
 if [ ! -e "${ANDROID_TOOLCHAIN}/$LD" ]; then
-    echo "ERROR: Failed to find Android ld. Please edit this script."
+    echo "ERROR: Failed to find Android ld. Please edit this script. When using NDK 22 or hihger make sure to set ANDROID_LD! (bfd, gold)"
```
MarcelRaad, line 374: Typo (hihger -> higher)

radioburst: Thanks for the input. -> fixed!

noloader: Thanks everyone. This fell off my radar. I want to look at this over the weekend.

noloader:
@radioburst,

The one comment I have is, `ANDROID_LD` should specify the linker name, and not a suffix. That is, you would do something like:

```bash
ANDROID_API="30" ANDROID_CPU="arm64-v8a" ANDROID_LD="ld.gold" source TestScripts/setenv-android.sh
```

There are two reasons. First, the non-Android makefile supports use of `LD=ld.gold`, and not a suffix like `.gold`.

Second, it looks like LLVM's `LLD` may be an [option in the future](https://android.googlesource.com/platform/ndk/+/refs/heads/ndk-release-r21/docs/BuildSystemMaintainers.md), and `lld` is going to be stand-alone and not a suffix. That is, this won't quite work with `LLD`:

```diff
    LD="arm-linux-androideabi-ld"
    if [ -z "${ANDROID_LD}" ]; then
        LD="arm-linux-androideabi-ld"
    else
        LD="arm-linux-androideabi-ld.${ANDROID_LD}"
    fi
```

If the linker name is fully specified, like `ld.gold` or `lld`, then we can do something like this and it should work with bfd, gold and lld.

```diff
    LD="arm-linux-androideabi-ld"
    if [ -z "${ANDROID_LD}" ]; then
        LD="arm-linux-androideabi-ld"
    else
        LD="arm-linux-androideabi-${ANDROID_LD}"
    fi
```

What do you think?

noloader:
@radioburst,

I think I have a second comment about this:

```diff
When using NDK 22 or higher make sure to set ANDROID_LD! (bfd, gold)"
```
If the user has to make a choice, I think we should supply a sane default so we don't break their workflow. We can provide the `ANDROID_LD` override, but we should also supply a sane default so "things just work."

According to [Build System Maintainers | Linkers](https://android.googlesource.com/platform/ndk/+/refs/heads/ndk-release-r21/docs/BuildSystemMaintainers.md#Linkers):

> Gold is used by default for most architectures, but BFD is used for AArch64 as Gold emits broken debug information for that architecture (see [Issue 70838247](https://issuetracker.google.com/70838247) for more details).

So this might look like

```diff
    LD="arm-linux-androideabi-ld"
    if [ "$ANDROID_API" -ge 22 ]; then
        LD="arm-linux-androideabi-ld.gold"
    fi
    if [ -n "${ANDROID_LD}" ]; then
        LD="arm-linux-androideabi-ld.${ANDROID_LD}"
    fi
```

What do you think?

noloader:
@radioburst,

I'm going to merge your PR, and then modify it as detailed so this works: `LD="arm-linux-androideabi-${ANDROID_LD}"`.

noloader:
@radioburst,

The changes tested Ok:

```
$ cp -p TestScripts/cryptest-android.sh .
$ cp -p TestScripts/setenv-android.sh .
$ ./cryptest-android.sh
...
armv7a ==> SUCCESS
armv7a : NEON ==> SUCCESS
aarch64 ==> SUCCESS
aarch64 : ASIMD ==> SUCCESS
aarch64 : AES ==> SUCCESS
aarch64 : PMULL ==> SUCCESS
aarch64 : SHA1 ==> SUCCESS
aarch64 : SHA2 ==> SUCCESS
x86 ==> SUCCESS
x86 : AES ==> SUCCESS
x86 : CLMUL ==> SUCCESS
x86 : SHA1 ==> SUCCESS
x86 : SHA2 ==> SUCCESS
x86_64 ==> SUCCESS
x86_64 : AES ==> SUCCESS
x86_64 : CLMUL ==> SUCCESS
x86_64 : SHA1 ==> SUCCESS
x86_64 : SHA2 ==> SUCCESS
```

But the tests are limited to build and link. We don't run the test suite.

radioburst:
@noloader 

I totally agree with you! Especially about the web people ;)
nah jokes aside! 
A sane default makes absolute sense. I actually thought about adding it but I completely forgot about it.
Thanks for the changes! This totally satisfies my needs.

noloader:
@radioburst,

We cut-in some more r22 changes at [Commit f799578860a1](https://github.com/weidai11/cryptopp/commit/f799578860a1).

Issue #735, opened 2018-11-08
Add ChaCha AVX2 implementation


noloader:
This ticket will track the cut-in of ChaCha AVX2 implementation. There a little more to this cut-in because AVX2 is new territory. I think we have most of the issues sorted out, including old Linux and Solaris with down-level compilers and linkers.

At the moment AVX2 is failing on one case. The case is VS2017, x64, Release build. AppVeyor result at [build 20159656](https://ci.appveyor.com/project/noloader/cryptopp/builds/20159656). It may be a similar problem as [Issue 649](https://github.com/weidai11/cryptopp/issues/649).

noloader:
Cleared over several commits.

Issue #649, opened 2018-04-30
Windows and incorrect results for AES when used on CPUs without AES-NI


rajkosto:
### Crypto++ Issue Report

Windows 7 - Windows 10 latest x64
Crypto++ 7.0.0
Compiled with Visual Studio 2017 latest version as of this writing (15.6.7) via cryptest.sln
(Building everything exactly the same with Visual Studio 2013 version 12.0.40629.00 Update 5 or Visual Studio 2015 Update 3, this bug does not happen)
After building either the dll or static library in Release x64 mode (Debug mode is fine, as are both Win32 builds), running dlltest.exe/cryptest.exe on a CPU with AES-NI instructions passes the AES tests, but when run on a CPU without AES-NI instructions the AES test fails (due to completely wrong output bytes). This can also be simulated on AES-NI CPUs by just setting g_AESNI = false; in DetectX86Features instead of reading it from CPUID.

noloader:
Thanks @rajkosto.

So I am clear, this is the summary:

* Visual Studio 2017 (15.6.7), Release, AESNI - PASS
* Visual Studio 2017 (15.6.7), Release, no AESNI - FAIL
* Visual Studio 2015, Release, AESNI - PASS
* Visual Studio 2015, Release, no AESNI - PASS
* Visual Studio 2013, Release, AESNI - PASS
* Visual Studio 2013, Release, no AESNI - PASS

Is that correct?

Also, could you download Sysinternal's [Coreinfo](https://docs.microsoft.com/en-us/sysinternals/downloads/coreinfo) and post the results?

This may be tough for me to work on because I may not be able to duplicate it.

-----

> This can also be simulated on AES-NI CPUs by just setting g_AESNI = false; in DetectX86Features instead of reading it from CPUID.

Setting just `g_AESNI = false` may miss the mark a bit. I think both `g_AESNI` and `g_hasCLMUL` should probably be either set or unset. But splitting them may make for a good test case, too.

noloader:
So it looks like encryption is OK. The problem appears to be on the decryption path for ECB and CBC mode. Below, notice the *"incorrectly decrypted"*. That is the second known answer test (with the first being encryption).

The results below were produced with the following change on Windows. Linux is OK:

```
$ git diff cpu.cpp
diff --git a/cpu.cpp b/cpu.cpp
index ef0ddf6..c5bc360 100644
--- a/cpu.cpp
+++ b/cpu.cpp
[...]

PR #1131, opened 2022-07-11
Remove obsolete workarounds for #649 and #735

chausner:
This removes the MSVC-specific workarounds added as part of #649 and #735. I tested with both the latest version of VS 2017 and VS2022 and could not reproduce the bug (I hardcoded the flag `g_hasAESNI` to false, built in Release mode, and ran the tests via `cryptest v`, as suggested in #649).

The workaround that had previously been applied leads to a massive slowdown of ChaCha, to a point where even the SSE2 version is much faster than the AVX2 implementation (so instead of disabling optimizations for AVX2, disabling the AVX2 variant altogether and falling back to SSE2 would have been the better workaround):

Before:

image

After, with optimizations enabled:

image

### DIFF
--- a/chacha_avx.cpp
+++ b/chacha_avx.cpp
@@ -36,13 +36,12 @@ extern const char CHACHA_AVX_FNAME[] = __FILE__;
 # define MAYBE_CONST const
 #endif
 
-// VS2017 and global optimization bug. TODO, figure out when
-// we can re-enable full optimizations for VS2017. Also see
+// VS2017 and global optimization bug. Also see
 // https://github.com/weidai11/cryptopp/issues/649 and
 // https://github.com/weidai11/cryptopp/issues/735. The
 // 649 issue affects AES but it is the same here. The 735
 // issue is ChaCha AVX2 cut-in where it surfaced again.
-#if (_MSC_VER >= 1910)
+#if (_MSC_VER >= 1910) && (_MSC_VER < 1916)
 # ifndef CRYPTOPP_DEBUG
 #  pragma optimize("", off)
 #  pragma optimize("ts", on)
--- a/rijndael.cpp
+++ b/rijndael.cpp
@@ -88,10 +88,9 @@ being unloaded from L1 cache, until that round is finished.
 #include "misc.h"
 #include "cpu.h"
 
-// VS2017 and global optimization bug. TODO, figure out when
-// we can re-enable full optimizations for VS2017. Also see
+// VS2017 and global optimization bug. Also see
 // https://github.com/weidai11/cryptopp/issues/649
-#if (_MSC_VER >= 1910)
+#if (_MSC_VER >= 1910) && (_MSC_VER < 1916)
 # ifndef CRYPTOPP_DEBUG
 #  pragma optimize("", off)
 #  pragma optimize("ts", on)


chausner:
In case you prefer to keep the workaround but disable it for unaffected compiler versions: I tested with the latest version of VS2017 which is MSVC compiler version 19.16. I am not sure what the exact version was when the bug had been fixed but it appears it was one of the updates between 19.10 and 19.16, so we could use
```c++
#if (_MSC_VER >= 1910) && (_MSC_VER < 1916)
```

Still, I suggest to just remove the workarounds completely as everyone should have updated to a fixed version by now, I would assume.

chausner:
Hmm, I can't even reproduce the issue with VS 2017 15.0 from https://visualstudio.microsoft.com/en/vs/older-downloads/. which bundles MSVC 16.10 (the version number that is currently hardcoded in the check):
```c++
#if (_MSC_VER >= 1910)
```
The exact versions I used for testing:
> VisualStudio/15.0.0-RTW+26228.102

and

> Microsoft (R) C/C++ Optimizing Compiler Version 19.10.25027 for x64

So the bug might have already been fixed in one of the 19.10.x patch versions or I am doing something wrong in trying to reproduce the issue.

chausner:
@noloader Any chance getting this performance regression fix reviewed and merged in the near future? Thanks! :)

chausner:
@noloader Thanks for your quick response!

> Perhaps you can carry a patch that unconditionally removes the workaround?

You mean that *conditionally* removes it? :) Unfortunately, I cannot reproduce the issue anymore with any still available version of the MSVC compiler. I think the compiler bug has simply been fixed a long time ago and there is no old version of the compiler available for download anymore that could reproduce it. So I am not exactly sure how to conditionally include the workaround if it is not clear which MSVC compiler versions are affected. Do you have any suggestion how to handle this?

In general, I don't think libraries need to try to stay compatible with every possible compiler bug that existed in the past, surely that would not be possible anyway. And in this case, I would assume many more users are nowadays negatively affected by the abysmal performance of ChaCha on Windows than those still using one of the old affected compiler versions. That's just my feeling but, of course, I respect your decision since you certainly know better what requirements users have on this library. I'd just like to find a pragmatic solution that solves the performance regression for users of modern compiler versions.

noloader:
@chausner,

Use this:

```
#if (_MSC_VER >= 1910) && (_MSC_VER < 1916)
```

> I don't think libraries need to try to stay compatible with every possible compiler bug that existed in the past, surely that would not be possible anyway. 

We do not govern ourselves like web projects, where you can't even get support for a 2 year-old operating system. You can still build and run Crypto++ apps on Fedora 1 and Ubuntu 4 with GCC 3.0; and Windows XP with Visual Studio .Net 2003. It is not hard to do.

> And in this case, I would assume many more users are nowadays negatively affected by the abysmal performance of ChaCha on Windows than those still using one of the old affected compiler versions.

No. The code **MUST** be correct, and the code **CAN** be fast. I won't put performance above correctness. If the code does not need to be correct, I can make it as fast as you like by commenting out the entire Salsa/ChaCha core block :)

chausner:
I have updated the MR with your suggestion, thanks. Please have another look.

In the end, it is up to library owners to decide how far they are willing to go to support older platforms/compilers. For some, backwards-compatibility is extremely important while for others, not at all. There is no right or wrong as different users have different requirements. Some are dependent on old software because they cannot or don't want to update, others update regularly and don't care about old versions. If support for old software is important to you or users of this library, then keeping the workaround with a conservative version range check is probably the best solution.

noloader:
Thank you very much.

noloader:
@chausner,

I tired to duplicate the problems but I could not. My version of VS2017 is MSC_VER=1916. I suspect problem was fixed some time between 1910 and 1916. Unfortunately we don't know when, so we are going to need to leave the fix in place over the wider range.

Thanks again for the help and testing it.

chausner:
You're welcome.

> I tired to duplicate the problems but I could not. My version of VS2017 is MSC_VER=1916.

Yes, I also could not reproduce it, not even with MSC_VER=1910 which is the version that was originally reported to have the bug. I suspect it might have been fixed with a patch release or it is not a bug in the compiler at all but a bug in the VC runtime which has their own updates.

chausner:
@noloader Can you estimate when these changes will make it into a new release of Crypto++? Is a new release planned anytime soon? 😊

I am using Conan to include this library in my project and it would require to set up my own port if I wanted to use the latest development version because only stable releases are usually available.

chausner-audeering:
I see there has been a new release 8.7 published with the fixes (even though they are not mentioned in the release notes). Thanks!

chausner:
I see there has been a new release 8.7 published with the fixes (even though they are not mentioned in the release notes). Thanks!

noloader:
@chausner,

I think removal of the workaround is causing an AppVeyor failure at https://ci.appveyor.com/project/noloader/cryptopp/builds/44561212/job/h34tffevqkctk0y8 .

We may need to enable the workaround for Visual Studio 2017.

chausner:
> I think removal of the workaround is causing an AppVeyor failure at https://ci.appveyor.com/project/noloader/cryptopp/builds/44561212/job/h34tffevqkctk0y8 .

Okay. Is it occurring in every CI build or only sporadically?

Interestingly, CI outputs
> hasSSE2 == 1, hasSSSE3 == 1, hasSSE4.1 == 1, hasSSE4.2 == 1, hasAVX == 1, hasAVX2 == 1, hasAESNI == 1, hasCLMUL == 1, hasRDRAND == 1, hasRDSEED == 0, hasSHA == 0, isP4 == 0

so AES-NI is available but #649 was originally reported to only occur on CPUs not supporting AES-NI. Maybe that assumption was wrong or it's a different bug.

I'll see if I can get my hands on the faulty binary from CI. Would be interesting to run it locally in a debugger to see where the error occurs exactly.

noloader:
@chausner,

Yep, it was the MSVC global optimization bug. When I changed to the following, the test stopped failing.

```c++
#if (_MSC_VER >= 1910) && (_MSC_VER <= 1916)
...
#endif
```

Previously we changed it to the following, which omitted the very last version of VS2017:

```c++
#if (_MSC_VER >= 1910) && (_MSC_VER < 1916)
...
#endif
```

Also see https://ci.appveyor.com/project/noloader/cryptopp/builds/44572833 .

chausner:
> I'll see if I can get my hands on the faulty binary from CI. Would be interesting to run it locally in a debugger to see where the error occurs exactly.

I looked some further into this, obtained the faulty binaries from AppVeyor and was able to reproduce the issue when I ran the binaries locally. 

- Not sure why VS2017 generates broken code on AppVeyor but not when building with the same VS version locally.
- I inspected the broken code in a disassembler and compared it with the correct one. There were differences in multiple places but I could not see any easy way how to workaround it.
- I tried to find a way to fix the bug without completely disabling global optimizations (e.g. by disabling inlining of some functions), in the hope that performance would not decrease as badly, but also had no success there.

I guess we'll just have to live with disabled optimizations on VS 2017. Luckily, I expect most users to be on newer VS versions nowadays and should thus not being affected.

noloader:
Also see [Commit 307ada58d618](https://github.com/weidai11/cryptopp/commit/307ada58d618). We re-enabled some global optimizations, but it was overly optimistic. We had to go back and tune it further in the 307ad commit.

Issue #1074, opened 2021-10-14
Can't build Crypto++ 8.6.0 for iOS in the M1 architecture


brutalkeso:
### Crypto++ Issue Report

```
crypto_version="CRYPTOPP_8_6_0"
function install_cryptopp_platform
{
    (
    set +u # Allow having undefined variables in setenv-ios.sh
    IOS_SDK="$1" IOS_CPU="$2" source ./TestScripts/setenv-ios.sh
    make -j"$NUM_PROC" -f GNUmakefile-cross
    make install -f GNUmakefile-cross PREFIX="$THIRDPARTY_INSTALL_PREFIX"
    mkdir -p "$THIRDPARTY_INSTALL_PREFIX/lib/crypto_$2"
    mv "$THIRDPARTY_INSTALL_PREFIX/lib/libcryptopp.a" "$THIRDPARTY_INSTALL_PREFIX/lib/crypto_$2"
    set -u
    )
}

function install_cryptopp
{
    echo "##teamcity[blockOpened name='cryptopp']"
    download_crypto
    local src_dir="cryptopp-$crypto_version"
    pushd "$src_dir"

    install_cryptopp_platform iPhoneOS arm64
    make clean

    install_cryptopp_platform iPhoneSimulator i386
    make clean

    local lib_arm64="$THIRDPARTY_INSTALL_PREFIX/lib/crypto_arm64/"
    local lib_i386="$THIRDPARTY_INSTALL_PREFIX/lib/crypto_i386/"

    lipo -create -output "$THIRDPARTY_INSTALL_PREFIX/lib/libcryptopp.a" "$lib_arm64/libcryptopp.a" "$lib_i386/libcryptopp.a"

    popd
    echo "##teamcity[blockClosed name='cryptopp']"
}
```

This works fine with `8_5_0`, but not with `8_6_0`. Then I get errors like:

```
clang++ -DNDEBUG -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS15.0.sdk -Wall -g2 -O3 -fPIC -arch arm64 -miphoneos-version-min=6 -stdlib=libc++ -fno-common -c dh.cpp
In file included from crc_simd.cpp:23:
./arm_simd.h:35:14: error: instruction requires: crc
    __asm__ ("crc32b   %w0, %w0, %w1   \n\t"
             ^
<inline asm>:1:2: note: instantiated into assembly here
[...]

PR #1154, opened 2022-09-13
Enable crc instructions when building for arm64 cpu (iOS)

jnavarrom:
I am not sure if this is the correct way to fix the compilation issue https://github.com/weidai11/cryptopp/issues/1074, but this works for me.

CRC instructions are enabled by default on ARMv8, but I don't know why it doesn't work when building for arm64, so I have added `mcrc` to `IOS_CXXFLAGS` if the CPU is arm64

> **-m[no-]crc**
> Enable or disable CRC instructions.
> 
> This option is used to indicate whether CRC instructions are to be generated. This only applies to the ARM architecture.
> 
> CRC instructions are enabled by default on ARMv8.
> 

https://clang.llvm.org/docs/UsersManual.html

### DIFF
--- a/TestScripts/setenv-ios.sh
+++ b/TestScripts/setenv-ios.sh
@@ -91,6 +91,7 @@ unset IOS_CFLAGS
 unset IOS_CXXFLAGS
 unset IOS_LDFLAGS
 unset IOS_SYSROOT
+unset CRC
 
 #########################################
 #####    Small Fixups, if needed    #####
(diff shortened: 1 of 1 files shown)


weidai11:
Thank you! I've replayed your commands and got the expected outputs, and then ran `git push origin linearized-history:master --force-with-lease` which seems to have succeeded.

StephanTLavavej:
Everything looks great now! Glad I could help. Now you can enable branch protections to require linear history, etc. if you haven't already.

PR #1170, opened 2022-10-16
Uninitialized member variable

Teebonne:
Fixes warning for uninitialized member variable


noloader:
`m_rounds` is set in `UncheckedSetKey`.

Also, this will break the compile of older C++ versions.

PR #1171, opened 2022-10-16
std::swap should not throw for C++11

Teebonne:
Adding noexcept with the macro CRYPTOPP_NO_THROW


noloader:
Sorry about the delay for this. I had to look it up to verify.

Thank you very much!

PR #1172, opened 2022-10-16
Uninitialized member variable

Teebonne:


noloader:
This is a scratch buffer that may (or may not) be used:

```diff
- byte m_buf[8];
+ byte m_buf[8] = {};
```

It does not need initialization. The writer will use as much (or little) of the buffer as needed.

Also, this will break the compile of older C++ versions.

PR #1173, opened 2022-10-24
Uninitialized member variable with macro

Teebonne:
See Pull request #1172.
Using a macro solve the issue of compatibility.


noloader:
Thanks @Teebonne.

I prefer not to do this.

PR #1187, opened 2023-01-23
Disable uncaught_exception support for C++ 20 and later

MarcelRaad:
`std::uncaught_exception` was removed from C++ 20.


noloader:
Thanks @MarcelRaad.

PR #1189, opened 2023-01-27
Add default target in makefile for mingw

Biswa96:


Biswa96:
If required I can try to add build and test procedure in CI with mingw toolchain.

ImLostPleaseHelpThx (approved):
Approved (Keep the base code in a file).

ImLostPleaseHelpThx:
> If required I can try to add build and test procedure in CI with mingw toolchain.

I have an operational script (have since I posted it).

noloader:
Hi Everyone.

Sorry about the late reply. I needed to look at some things because the DLL was traditionally the FIPS DLL. The FIPS DLL is going away. Also see .

But I see this in the `GNUmakefile` when examining the `cryptopp.dll` recipe and the `DLLOBJS` prereq:

```
# In Crypto++ 5.6.2 these were the source and object files for the FIPS DLL.
# Since the library is on the Historical Validation List we add all files.
# The 5.6.2 list is at https://github.com/weidai11/cryptopp/blob/789f81f048c9.
DLLSRCS := $(SRCS)
DLLOBJS := $(DLLSRCS:.cpp=.export.o)
DLLOBJS := $(DLLOBJS:.S=.export.o)
```

I don't use MinGW, so please forgive my ignorance... Does `cryptopp.dll` actually work on Windows using MinGW? Are all the classes there? Does `cryptest.exe vv` run and pass its self tests?

Biswa96:
cryptest.exe does not link with the import library libcryptopp.dll.a. So, I can not test the shared library. But cryptest.exe links with static library and all tests passed. I added these changes to make sure libraries can be built and installed in mingw like other platforms. Then investigate the shared library issue if possible.

PR #1232, opened 2023-09-28
Update osrng.cpp

Kronix69:
Added the possibility of using Microsoft's default RNG provider, removes the need to acquire a provider (which solves the problem of not having an available provider and also saves resources)


noloader:
Thanks @Kronix69.

Where is `CRYPTOPP_DEFAULT_AES_PROVIDER` configured? I only see it used in the change request.

Kronix69:
I momentarily left it for so the user can enable it if they desire and so if you guys believed it to be better to enable it by default

Kronix69:
We could completely remove the need for a provider, I haven't encountered any issues with using the default provider so we could just erase the whole acquiring of a provider

Kronix69:
Hey @noloader, any status on merging?

wyattoday:
`BCRYPT_RNG_ALG_HANDLE` is only defined (and usable) on Windows 10 and newer. So, this will fail if targeting older versions of Windows.

You could [do a runtime detection of the version](https://stackoverflow.com/a/69121369/124805) and optionally create the `MicrosoftCryptoProvider()` (and use that handle) or use the `BCRYPT_RNG_ALG_HANDLE` depending on what version of Windows the end-user is on.

Kronix69:
> `BCRYPT_RNG_ALG_HANDLE` is only defined [(and usable) on Windows 10 and newer](https://learn.microsoft.com/en-us/windows/win32/seccng/cng-algorithm-pseudo-handles). So, this will fail if targeting older versions of Windows.
> 
> You could conditionally define `BCRYPT_RNG_ALG_HANDLE` and [do a runtime detection of the version](https://stackoverflow.com/a/69121369/124805) and optionally create the `MicrosoftCryptoProvider()` (and use that handle) or use the `BCRYPT_RNG_ALG_HANDLE` depending on what version of Windows the end-user is on.

Done, I instead used the WINVER define which specifies the Windows version to check for Windows 10 and higher.

Kronix69:
Any status on the merge?

Kronix69:
Nevermind I realised the issue with static checking, will devise a way to runtime check

PR #1266, opened 2024-03-18
Fixing an infinite loop scenario in case of memory exhaustion

jvanegue:
Throwing an exception after 10 attempts rather than trying to allocate indefinitely (which could be infinite)

### DIFF
--- a/allocate.cpp
+++ b/allocate.cpp
@@ -39,6 +39,7 @@ void CallNewHandler()
 
 void * AlignedAllocate(size_t size)
 {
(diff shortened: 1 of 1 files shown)


jvanegue:
Any update on the above? This file has two distinct locations where this problem may happen.
Thank you

jvanegue:
Hello! Have you considered accept this PR? This introduces a fuel pattern for your allocations to avoid potential infinite loops. Thanks.

PR #1291, opened 2024-09-16
Fix "unreachable code" warning

irwir:
Visual C++ emits this warning for "return false;" line if one of the previous conditions has been met.


cpu.cpp
```diff
@@ -857,8 +857,9 @@ inline bool CPU_QueryARMv7()
 #elif defined(_WIN32) && defined(_M_ARM64)
 	// Windows 10 ARM64 is only supported on Armv8a and above
 	return true;
-#endif
+#else
```
MarcelRaad, line 860: The Linux and Android cases above are missing a `return` now in case the run-time conditions are not met.

irwir:
Thanks, @MarcelRaad. The code has been updated.

PR #1306, opened 2025-03-21
Fix compile break when building with clang-cl

ChristophStrehle:
Fix a compile break when building with Clang and the Microsoft STL


ChristophStrehle:
this would break builds for VS < 2022

Issue #1311, opened 2025-04-04
Compile break when building with Microsoft STL using a different compiler then msvc


ChristophStrehle:
When building with the Microsoft STL and a different compiler then the msvc I get the following compile break in std::vector:

```
1>..\MSVC\14.43.34808\include\vector(656,38): error : no matching conversion for static_cast from '_Alty' (aka 'AllocatorWithCleanup<unsigned char, false>') to '_Rebind_alloc_t<_Alty, _Container_proxy>' (aka 'AllocatorWithCleanup<std::_Container_proxy, false>')
1>..\MSVC\14.43.34808\include\vector(656,38): error :   656 |         _Mypair._Myval2._Alloc_proxy(_GET_PROXY_ALLOCATOR(_Alty, _Getal()));
1>..\MSVC\14.43.34808\include\vector(656,38): error :       |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1>..\MSVC\14.43.34808\include\xmemory(1504,42): message : expanded from macro '_GET_PROXY_ALLOCATOR'
1>..\MSVC\14.43.34808\include\xmemory(1504,42): message :  1504 | #define _GET_PROXY_ALLOCATOR(_Alty, _Al) static_cast<_Rebind_alloc_t<_Alty, _Container_proxy>>(_Al)
1>..\MSVC\14.43.34808\include\xmemory(1504,42): message :       |                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1>..include\crypto++/xed25519.h(284,5): message : in instantiation of member function 'std::vector<unsigned char, CryptoPP::AllocatorWithCleanup<byte>>::vector' requested here
1>..include\crypto++/xed25519.h(284,5): message :   284 |     ed25519_MessageAccumulator() {
1>..include\crypto++/xed25519.h(284,5): message :       |     ^
1>..include\crypto++\secblock.h(187,7): message : candidate constructor (the implicit copy constructor) not viable: no known conversion from 'AllocatorWithCleanup<unsigned char, [...]>' to 'const AllocatorWithCleanup<std::_Container_proxy, [...]>' for 1st argument
1>..include\crypto++\secblock.h(187,7): message :   187 | class AllocatorWithCleanup : public AllocatorBase<T>
[...]

PR #1315, opened 2025-04-15
Fix compile break when building with the Microsoft STL

ChristophStrehle:
Fixed issue: #1311

### DIFF
--- a/config_ver.h
+++ b/config_ver.h
@@ -84,9 +84,28 @@
 # define CRYPTOPP_INTEL_VERSION (__INTEL_COMPILER)
 #endif
 
-#if defined(_MSC_VER) && !defined(__clang__)
-# undef CRYPTOPP_LLVM_CLANG_VERSION
-# define CRYPTOPP_MSC_VERSION (_MSC_VER)
+#if defined(_MSC_VER)
+#include <yvals_core.h>		// needed for _MSVC_STL_VERSION
+# if !defined(__clang__)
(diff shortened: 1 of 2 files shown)


irwir:
Is it necessary to define STL version?
If not, the changes could be as short as:

```
#if defined(_MSVC_STL_VERSION) // VS2017 (14.1) and above
# define CRYPTOPP_MSSTL
#elif defined(_CPPLIB_VER) && _CPPLIB_VER >= 503 // VS2008 (also 9.0)
# define CRYPTOPP_MSSTL
#endif

```
and
`#ifdef CRYPTOPP_MSSTL`

ChristophStrehle:
Hello irwi,
the suggestion you made would resolve this specific issue. But it would also define `CRYPTOPP_MSC_VERSION` when building with clang-cl. I think the intention of a89a27b72ced9a38a0a7d8548ee8b198c7340b93 was to have `CRYPTOPP_MSC_VERSION` only defined when building with MSVC but not when building with clang-cl.

irwir:
Have you read carefully the issue 147 referenced in the mentioned PR?
The "attempt to unbind clang from others" mostly failed, because on Windows clang is not self-sufficient and works on top of an other compiler.

If definition of `CRYPTOPP_MSC_VERSION` was undesirable - even though it did not break anything, then nothing would be simpler than reverting to original `#if _MSC_VER >= 1500` check in `secblock.h`.
Or, in more formal way, see https://github.com/weidai11/cryptopp/pull/1315#issuecomment-2966171982

PR #1324, opened 2025-05-27
Fix arm crypto logic issue in config_asm.h

antznin:
The current logic for enabling or disabling the `CRYPTOPP_ARM_*_AVAILABLE` macros seems wrong, because we do an OR between all of the conditions following `defined(__ARM_FEATURE_CRYPTO)`. Fix the logic so that the `CRYPTOPP_ARM_*_AVAILABLE` macros are set if __ARM_FEATURE_CRYPTO is set _AND_ any of the following condition is true.

### DIFF
--- a/config_asm.h
+++ b/config_asm.h
@@ -278,9 +278,9 @@
 // Requires GCC 4.8, Clang 3.3 or Visual Studio 2017
 #if !defined(CRYPTOPP_ARM_AES_AVAILABLE) && !defined(CRYPTOPP_DISABLE_ARM_AES)
 # if defined(__aarch32__) || defined(__aarch64__) || defined(_M_ARM64)
-#  if defined(__ARM_FEATURE_CRYPTO) || (CRYPTOPP_GCC_VERSION >= 40800) || \
+#  if defined(__ARM_FEATURE_CRYPTO) && ((CRYPTOPP_GCC_VERSION >= 40800) || \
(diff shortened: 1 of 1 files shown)


antznin:
> > The current logic for enabling or disabling the `CRYPTOPP_ARM_*_AVAILABLE` macros seems wrong
> 
> OR'ed condition contains a list of known and allowed compilers; otherwise `__ARM_FEATURE_CRYPTO` could be used to force  enabling of `CRYPTOPP_ARM_*_AVAILABLE` macros.

That was the case before my patch, yes.

Now after my modification, a known and allowed compiler must be used _and_ `__ARM_FEATURE_CRYPTO` must be set.

So I'm not sure I understand your comment.

irwir:
> Now after my modification, a known and allowed compiler must be used _and_ `__ARM_FEATURE_CRYPTO` must be set.

Incorrect interpretation of code logic.
Should cryptopp need a single on/off switch here, then `__ARM_FEATURE_CRYPTO` would be checked before other conditions. In fact, all three blocks you suggested to change could have been placed under one check.

antznin:
> Should cryptopp need a single on/off switch here, then `__ARM_FEATURE_CRYPTO` would be checked before other conditions. In fact, all three blocks you suggested to change could have been placed under one check.

Sure, I can change the code so `__ARM_FEATURE_CRYPTO` is checked first, then the three compilers. This also makes sense to me. But I'm not sure that what we want after your comments.

irwir:
This PR goes against portability, because any non-listed compiler would require source code changes.

PR #1340, opened 2025-11-13
Fix ESIGN memcpy null pointer and clang22 fortify warning (#1338)

### LINKED ISSUE #1338: Fortify warning: memcpy bound exceeds object size in esign.cpp when building with clang22
Hello maintainers,

When building Crypto++ with clang22 (flags: -DNDEBUG -g2 -O3 -fPIC -fno-devirtualize -pthread -pipe), I observed a fortify/stringop-overflow warning from `memcpy` at compilation of esign.cpp. I would appreciate your advice or triage for this warning. 

---

**Warning excerpt:**
```
In file included from /usr/include/string.h:548,
                 from /usr/include/c++/14/cstring:43,
                 from stdcpp.h:63,
                 from cryptlib.h:106,
                 from esign.h:10,
                 from esign.cpp:11:
In function 'void* memcpy(void*, const void*, size_t)',
    inlined from 'virtual void CryptoPP::InvertibleESIGNFunction::GenerateRandom(CryptoPP::RandomNumberGenerator&, const CryptoPP::NameValuePairs&)' at esign.cpp:115:14:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:33: warning: 'void* __builtin_memcpy(void*, const void*, long unsigned int)' specified bound 18446744073709551612 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
   29 |   return __builtin___memcpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   30 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:33: warning: 'void* __builtin_memcpy(void*, const void*, long unsigned int)' specified bound 18446744073709551612 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
```

---

**Context & Source:**
- Source file: [`esign.cpp` @ commit 843d74c7](https://github.com/weidai11/cryptopp/blob/843d74c7c97f9e19a615b8ff3c0ca06599ca501b/esign.cpp)
- Potentially relevant code:
```c++
if (param.GetValue("Seed", seedParam))
{
    seed.resize(seedParam.size() + 4);
    std::memcpy(seed + 4, seedParam.begin(), seedParam.size());
    // ...
}
```
(see full file: [`esign.cpp`](https://github.com/weidai11/cryptopp/blob/843d74c7c97f9e19a615b8ff3c0ca06599ca501b/esign.cpp))

- Pipeline reproduction: [`cryptobench-cpp` build log here](https://github.com/scc-tw/cryptobench-cpp/actions/runs/19099030093/job/54566151595)
- Compilation flags used: clang22  -DNDEBUG -g2 -O3 -fPIC -fno-devirtualize -pthread -pipe

---

**Notes:**
- I have not investigated the root cause.
- I am not proposing a patch at this time; this is a tracking/triage report.

**Steps to reproduce:**
1. Build Crypto++ with clang22 and the flags listed above.
2. Observe the warning in the compilation output for esign.cpp.

Many thanks for your attention!

scc-tw:
- Fixing the immediate UBSan bug (null pointer to memcpy when the size is 0) 
- Quieting the Clang/Fortify warning at `InvertibleESIGNFunction::GenerateRandom(RandomNumberGenerator &rng, const NameValuePairs &param)`

Fixes #1338

### DIFF
--- a/esign.cpp
+++ b/esign.cpp
@@ -115,7 +115,11 @@ void InvertibleESIGNFunction::GenerateRandom(RandomNumberGenerator &rng, const N
 			throw InvalidArgument("InvertibleESIGNFunction::GenerateRandom: buffer overflow");
 
 		seed.resize(seedParam.size() + 4);
-		std::memcpy(seed + 4, seedParam.begin(), seedParam.size());
+		// Help static analyzer verify bounds (Issue #1338)
+		CRYPTOPP_ASSERT(seed.size() >= seedParam.size() + 4);
+		// Guard against null pointer when size is 0 (UBSan)
+		if (seedParam.size() > 0)
+			std::memcpy(seed + 4, seedParam.begin(), seedParam.size());
 
 		PutWord(false, BIG_ENDIAN_ORDER, seed, (word32)0);
 		m_p.GenerateRandom(rng, CombinedNameValuePairs(primeParam, MakeParameters("Seed", ConstByteArrayParameter(seed))));


Coralesoft:
@irwir, thanks for the thorough review. You’ve raised valid technical points:

**Re: pointer arithmetic**, you’re correct that `seed.data() + 4` is still pointer arithmetic. The benefit is that using an explicit `.data()` pointer helps Clang/Fortify track buffer bounds in static analysis, which reduces false positives in practice. The runtime semantics are identical, but the tooling behaviour improves.

**Re: `CRYPTOPP_ASSERT`**, agreed it’s a no-op in release builds when `NDEBUG` is defined (trap.h:101). In practice, the runtime guard in this code path is the check in `esign.cpp` (around line 114):

```cpp
if (seedParam.size() > seed.ELEMS_MAX - 4)
    throw InvalidArgument(...);
```

The assert is there as supplementary debug-time validation.

**Re: `ELEMS_MAX`**, this is your strongest point. `ELEMS_MAX = SIZE_MAX/sizeof(T)` can exceed `malloc`’s practical limits (often around `PTRDIFF_MAX` on most systems). While the guard above prevents issues in this particular code path, you’re right that the underlying concern lives in `secblock.h` rather than in this function.

**Proposal**, my view is that @scc-tw's PR still has value on its own by:
- Fixing the immediate UBSan bug (null pointer to `memcpy` when the size is 0), and  
- Quieting the Clang/Fortify warning at this call site.

I’d suggest we treat this as a targeted fix and track the `ELEMS_MAX` architecture question separately, if you think it’s worth pursuing, it might make sense to open a follow-up issue for the `ELEMS_MAX` concern you’ve identified in `secblock.h`. I’d be happy to contribute to the discussion.

Thoughts?

scc-tw:
@irwir Thank you again for your thoughtful comments.

I agree that the `ELEMS_MAX` topic and realistic allocation limits belong to a broader design discussion around `SecBlock`, and are separate from this localized UBSan/Fortify issue. This PR is only meant to address the concrete UBSan runtime error and the Fortify/Clang diagnostic at this specific call site, without changing the behavior of `GenerateRandom` for non-empty seeds.

To properly address the `ELEMS_MAX` concern, I will open a separate issue so we can track it and work toward a clean, general solution there. If you have any suggestions for that discussion, I would really appreciate your input.

PR #1351, opened 2026-05-06
windows ARM64 build enablement using visual studio

pdeep854:
Added build configuration changes to enable Windows ARM64 builds using the Visual Studio build system.


noloader:
I would like to add this to the library, but I don't have the test rigs to test it.  I'm going to wait until someone like David tests it.

config_asm.h
irwir, line 364: Looks overcomplicated, the same conditional `define` as for NEON could have been used instead.
```
#  if !defined(_M_ARM64)
#   define CRYPTOPP_ARM_ACLE_HEADER 1
#  endif
```

pdeep854: addressed the review comments. please take another look.

Issue #1352, opened 2026-05-20
Ed25519 verification accepts non-canonical signatures with S >= L


Wowblk:
# Ed25519 verification accepts non-canonical signatures with `S >= L`

Hi maintainers,

I found that Crypto++ 8.9.0 accepts non-canonical Ed25519 signatures where the encoded scalar `S` is not less than the Ed25519 subgroup order `L`.

For Ed25519, the signature is encoded as `R || S`, where `S` is a 32-byte little-endian integer. RFC 8032 requires verification to reject signatures with `S >= L`. However, Crypto++ accepts a signature with `S = L`, treating it like the equivalent scalar `S = 0` in the group equation.

## Impact

This allows multiple byte encodings of the same logical signature to verify successfully. Higher-level systems that treat signature bytes as canonical can be affected, for example:

- signature deduplication or replay caches keyed by raw signature bytes
- audit or logging systems expecting a unique signature encoding
- blocklists or allowlists over serialized signatures
- cross-implementation workflows where stricter Ed25519 implementations reject the same signature that Crypto++ accepts

This is similar in shape to public-key canonicalization issues, but the non-canonical encoding is in the signature scalar `S`.

## Proof of Concept

The following PoC uses:

- public key: canonical identity point, `y = 1`
- signature `R`: canonical identity point
- signature `S`: subgroup order `L`

Since `S = L` is not a canonical Ed25519 scalar encoding, verification should fail.

```cpp
#include <cryptopp/xed25519.h>
#include <cstring>
#include <iostream>

int main() {
    using CryptoPP::byte;

    const byte message[] = {'t','e','s','t'};

    byte public_key[32] = {};
    public_key[0] = 1; // identity public key

    byte signature[64] = {};
    signature[0] = 1; // R = identity

    const byte order_l[32] = {
        0xed,0xd3,0xf5,0x5c,0x1a,0x63,0x12,0x58,
        0xd6,0x9c,0xf7,0xa2,0xde,0xf9,0xde,0x14,
        0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
        0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x10
    };

    std::memcpy(signature + 32, order_l, sizeof(order_l)); // S = L

    CryptoPP::ed25519Verifier verifier(public_key);
    bool accepted = verifier.VerifyMessage(
        message, sizeof(message), signature, sizeof(signature));

    std::cout << std::boolalpha
[...]

PR #1355, opened 2026-05-26
Reject Ed25519 signatures with S >= L (GH #1352)

Coralesoft:
## Summary

Addresses the signature-scalar part of GH #1352.

The Ed25519 verifier accepts non-canonical signatures where `S >= L` because the existing check only masks the top three bits of `S`, leaving the gap `L <= S < 2^253`. A valid signature changed from `S'` to `S' + L` still verifies because the equation holds modulo the subgroup order.

Both the Donna verifier and the NaCl C API verifier had this defect. Both are patched here in separate commits.

## What changed

- **Donna:** added `ed25519_scalar_is_canonical` to `donna_64.cpp` and `donna_32.cpp`, called at the start of both `ed25519_sign_open_CXX` overloads. The old `(RS[63] & 224)` top-bit mask was removed since the full order check covers it.
- **NaCl:** added `scalar_is_canonical` to `tweetnacl.cpp`, called in `crypto_sign_open` before `unpackneg`. Reused the existing `L` constant at file scope.

### DIFF
--- a/donna_32.cpp
+++ b/donna_32.cpp
@@ -2028,6 +2028,22 @@ ed25519_sign(const byte* message, size_t messageLength, const byte secretKey[32]
     return ed25519_sign_CXX(message, messageLength, secretKey, publicKey, signature);
 }
 
+/* Reject S >= L (RFC 8032 canonical encoding check). */
+inline bool
+ed25519_scalar_is_canonical(const byte S[32]) {
+    static const byte L[32] = {
+        0xed,0xd3,0xf5,0x5c,0x1a,0x63,0x12,0x58,
+        0xd6,0x9c,0xf7,0xa2,0xde,0xf9,0xde,0x14,
+        0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
+        0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x10
+    };
+    for (int i = 31; i >= 0; --i) {
+        if (S[i] < L[i]) return true;
+        if (S[i] > L[i]) return false;
+    }
+    return false;
+}
+
 int
 ed25519_sign_open_CXX(std::istream& stream, const byte pk[32], const byte RS[64]) {
 
@@ -2038,7 +2054,7 @@ ed25519_sign_open_CXX(std::istream& stream, const byte pk[32], const byte RS[64]
     bignum256modm hram, S;
     byte checkR[32];
 
-    if ((RS[63] & 224) || !ge25519_unpack_negative_vartime(&A, pk))
+    if (!ed25519_scalar_is_canonical(RS + 32) || !ge25519_unpack_negative_vartime(&A, pk))
         return -1;
 
     /* hram = H(R,A,m) */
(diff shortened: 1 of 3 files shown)


noloader:
Thanks @Coralesoft.

> Both the Donna verifier and the NaCl C API verifier had this defect. Both are patched here in separate commits.

I'd be interested to know what Bernstein and Moon think about the change.  I'm thinking they already accounted for it by masking the high order byte.  Have you performed any research?

Jeff

Coralesoft:
Hi Jeff,

It's been a while, but I found my notes. You are right about donna: floodyberry's tree at HEAD still has only the `RS[63] & 224` mask and no full `S < L` check.

The catch is that the mask and canonicity are different bounds. `& 224` enforces `S < 2^253`, while canonical encoding needs `S < L`, and `L` is only slightly above `2^252`. Every scalar encoding in `[L, 2^253)` passes that mask, and that is where `S + L` lands for all but roughly a `2^-127` fraction of canonical scalars.

Chalkias, Garillot and Nikolaenko surveyed the major Ed25519 implementations in "Taming the many EdDSAs" (https://eprint.iacr.org/2020/1244.pdf, SSR 2020) and put donna in the same category as ref10:

> The exceptions are ed25519-java, TweetNacl, python-ed25519, ed25519-donna, and ref10, the latter two of which only perform the incomplete fail fast check (as shown in Listing 1.1 line#4), rather than a full check of its size.

Their Listing 1.1 shows how both masks can be used as fast paths around the full comparison:

```text
if s_bytes[31] & 240 == 0 { true }        // succeed fast
else if s_bytes[31] & 224 != 0 { false }  // fail fast
else { full_s_canonicity_check(s_bytes) }
```

Current libsodium uses the `& 240` succeed-fast check before calling `sc25519_is_canonical`. Its `ED25519_COMPAT` branch retains the older `& 224`-only behaviour.

Here is what happens on upstream commit `78242590`. Using RFC 8032 test vector 1 and then the same signature with `L` added to `S`, both signatures verify:

```text
public    d75a980182b10ab7d54bfed3c964073a0ee172f3daa62325af021a68f707511a
message   (empty)

original  e5564300c360ac729086e2cc806e828a84877f1eb8e5d974d873e06522490155
          5fb8821590a33bacc61e39701cf9b46bd25bf5f0595bbe24655141438e7a100b

malleated e5564300c360ac729086e2cc806e828a84877f1eb8e5d974d873e06522490155
          4c8c7872aa064e049dbb3013fbf29380d25bf5f0595bbe24655141438e7a101b
```

The last byte of `S` goes from `0x0b` to `0x1b`, `0x1b & 224` is still zero, and `[S+L]B == [S]B` leaves the equation unchanged. Two signatures are accepted by current master for one key and one message, though the second is invalid under RFC 8032 section 5.1.7, which requires `0 <= S < L`.

The mask is not useless, for what it is worth. It does reject `S` far above `L`, which is why the paper records donna and ref10 rejecting its vector 7 while accepting vector 6. This `S + L` value sits in the band between them.

Their archived test-vector repository is at https://github.com/novifinancial/ed25519-speccheck if you want to check this independently of my patch. Vectors 6 and 7 are the `S`-out-of-bounds cases, and the repo publishes a pass/fail table across the libraries they tested.

Col

